PR #238, opened 2015-02-16
Add episode ID to response object

bricker:
This would be good to have since there's an endpoint to fetch an episode by its numerical ID but no way to find the ID through the API. This isn't blocking me on anything, so no rush or obligation.


sdillingham:
Looks good, merging.

bricker:
Thanks, I'll update the documentation too.

bricker:
I'm also putting together a list of other suggestions for the API, none of which are blocking me in any way for the app but worth considering: https://gist.github.com/bricker/73c973b83ae70550b108

PR #249, opened 2015-04-06
Mobile push notifications

sdillingham:
Splits Parse push notifications out into two channels, one for iPad (breakingNews) and one for iPhone (listenLive). When the breaking news alert_type is a "Listen Live" event, the notification will be pushed to both iPhone and iPad apps; when the alert_type is either "Breaking News" or "Happening Now" push notifications will only be sent to the iPad app. The idea behind this is that the iPhone app is a listening-focused experience and so only alerts related to the live stream should push to that product.


ewr:
Looks like you accidentally included a Redis dump.

sdillingham:
derp. I thought I had added that to my .gitignore. I'll remove it now

Issue #243, opened 2015-03-11
Project should have usable default config files


ewr:
Currently, the repo doesn't have checked-in versions of config files like `database.yml`, `secrets.yml` or `api_config.yml`. That makes startup harder for a new developer.

SCPRv4 should contain sensible default versions of all config files needed for a developer to get the project up and running.

Obviously that shouldn't include sensitive API keys, etc. If any sensitive API keys are necessary to get the app up and running, they should be well-documented in the README.

ewr:
Closed via #253

PR #253, opened 2015-05-04
Rails 4.2 and Cleaned Up Configuration

ewr:
Upgrade to Rails 4.2 and clean up how we do app configuration.

* `app_config.yml` and `api_config.yml` become `secrets.yml`
* Add a developer `database.yml`
* Add a developer `newrelic.yml`
* Add `config/environments/development.rb`

Tests pass and it seems to work well deployed to staging, so we should be able to merge this soon.


ewr:
This would close #243.

Issue #250, opened 2015-04-27
Podcast feeds should support a "consumer" param


sdillingham:
Currently our podcast feeds expose two parameters (`context` and `via`) when constructing audio enclosure URLs that are used for tracking purposes. See:

https://github.com/SCPR/SCPRv4/blob/master/app/views/podcasts/podcast.xml.builder#L51-L54

`via` is used to ascertain what channel the audio was published from (e.g., "podcast," "website," "api"); `context` is used to reliably pass in the entity/show the audio belongs to (e.g., "airtalk," "the-frame," "latest-news").

We need a third parameter to track specific 3rd-party consumers of our feeds so we can know with confidence which aggregators are driving the greatest amount of podcast downloads. The goal would be to allow `?consumer=xxx` to be appended to any podcast URL (e.g `http://scpr.org/podcasts/the-frame?consumer=iheartradio`), and have the podcast feed generator key off that param and inject it into all audio enclosure URLs for the given feed.

ewr:
Closed via #257

ewr:
So... we missed the caching...

ewr:
Fixed via #259.

PR #257, opened 2015-05-05
Podcast consumer query param

Ravenstine:
In reference to: https://github.com/SCPR/SCPRv4/issues/250

A third party can now add a ```consumer``` key to the get params for a podcast; the value of that consumer key will show up in every item enclosure url.


ewr:
:+1: 

We just need to wait until the config changes get merged before merging this.

Ravenstine:
Thanks for the code review; I will make those changes now.

PR #259, opened 2015-05-06
Include the consumer value in the cache key for the podcast xml builder.

Ravenstine:


ewr:
Looks good to me. :shipit:

PR #260, opened 2015-05-07
Add coverage timestamps to tags and update them on content save

Ravenstine:
One thing that worries me a little is schema.rb; there are a lot of changes that add extra options to several lines.  It doesn't appear to have made any significant changes like removing columns, so it may be okay.  If it looks like there's a problem, I will fix the PR.


ewr:
I have a conceptual question: In this, you're re-computed earliest and latest dates every time a tagging is updated. I don't believe we need to do that. 

Each time we touch a tag, we just need to see if our published_at date is earlier or later than those already on the tag. That gives us a much more optimized path, since all we need loaded is our content and our tag.

The exception is when we untag something, which I'm not even sure is handled in the code right now. In that case we would want to ask the tag for its first and last content without the piece being removed. The tag has an existing way to do that (`taggables`), so we don't need to build that query into the association.

Ravenstine:
I misunderstood the intent of what we were wanting to achieve.  I have changed the pull request; anything associated with a tag and tries to call touch_tags after save will give its published_at timestamp its tags so that they can update either began_at or most_recent at depending on whether the published_at is greater than or less than either of those fields.

Should I now account for when something is untagged?

db/migrate/20150505231843_add_coverage_timestamps_to_tags.rb
```diff
@@ -0,0 +1,13 @@
+class AddCoverageTimestampsToTags < ActiveRecord::Migration
+  def up
+  	add_column :tags, :began_at, :datetime, index: true, default: Time.now
```
ewr: I don't think we need to index these columns. I'm also not sure whether that default will apply... I don't think the schema is actually used in the model itself.

db/migrate/20150505231843_add_coverage_timestamps_to_tags.rb
```diff
@@ -0,0 +1,13 @@
...
+  	add_column :tags, :most_recent_at, :datetime, index: true, default: Time.now
+  	Tag.all.each do |tag|
```
ewr: Generally I've seen best practices be to do these steps in two migrations. Create the fields in one, do the data update in the next one.

Rails also keeps a cache of columns, so in that second migration you need to use:

```ruby
Tag.reset_column_information
```

db/migrate/20150505231843_add_coverage_timestamps_to_tags.rb
```diff
@@ -0,0 +1,13 @@
...
+			tag.update began_at: tag.articles.last.created_at, most_recent_at: tag.articles.first.created_at
```
ewr: We should be using the content's `public_datetime`, not `created_at`.

db/migrate/20150505231843_add_coverage_timestamps_to_tags.rb
Ravenstine: True; how about using a before_save to enforce the default?

ewr: I think you actually don't want a default. Imagine you're creating a tag to scoop up some old stories. If your default end tag was now, you would never get a proper end tag from the old content. 

We may just need to guard that display in the view template instead.

app/models/tag.rb
```diff
@@ -14,4 +14,26 @@ def taggables(options={})
   def articles(options={})
     taggables(options)
   end
+
+  def update_timestamps published_at
+    updates = {}
+
+    if began_at && (published_at < began_at) || began_at.nil?
```
ewr: Couldn't these tests just be `began_at.nil? || (published_at < began_at)`?

app/models/tag.rb
```diff
@@ -14,4 +14,26 @@ def taggables(options={})
...
+      updates[:most_recent_at] = published_at
+    end
+
+    update(updates) if updates.any?
+    self
+  end
+
+  private
+
+  def connection
```
ewr: Is this getting used for something, or is it leftover from a different iteration?

db/migrate/20150511200111_populate_tag_coverage_timestamps.rb
```diff
@@ -0,0 +1,13 @@
+class PopulateTagCoverageTimestamps < ActiveRecord::Migration
+  def up
+    Tag.reset_column_information
+    Tag.all.each do |tag|
+      articles = tag.articles
```
ewr: `Tag.articles` is a ContentBase search (ES) rather than the underlying association. That means it's going to apply auto-pagination here. 

Test this, but I think you would want to do two queries:

Most Recent: `tag.articles(order:"public_datetime desc",limit:1)`
Oldest: `tag.articles(order:"public_datetime asc",limit:1)`

(Note that ES is funny about requiring `desc` and `asc` to be lowercase)

ewr:
I think this should be good to go once those new comments are addressed. It doesn't account for date recalculation when a tag is removed right now, but I don't see that as a particularly big worry.

app/models/tag.rb
Ravenstine: It's being used by the tags_association concern(line 31), which formerly called #touch on each of its tags now calls #update_timestamps with the published_at date as an argument.

ewr:
LGTM. I'd say let's merge into master and then test deploy it to staging to make sure the migration is happy.

PR #264, opened 2015-05-08
External episode cleaning

Ravenstine:
Both adds a sweeper to remove expired episodes as well as preventing expired episodes [that still have records] from being displayed for a program.

The expiry is determined by the 'days_to_expiry' column in the external_programs table.  A Resque job is added to destroy_all expired records once a day at 1am(just picked it for the heck of it).


ewr:
Let's think about simplifying this a bit before we merge it. Whenever possible, I want to do things the simple way rather than the clever way.

Help me understand the intent of the two new concerns and the MissingEpisode model... Why are we better off with a MissingEpisode model rather than just letting episode be `nil` and handling that case appropriately?

I also think we want to avoid putting view rendering information in the models, so let's try and refactor the controller refactors to handle that themselves.

app/models/external_program.rb
```diff
@@ -125,6 +128,33 @@ def sync
     self.importer.sync(self)
   end
 
+  def expired_episodes
+    if days_to_expiry
+      external_episodes.where('DATE(?) >= DATE_ADD(created_at, INTERVAL ? DAY)', Time.now, days_to_expiry)
```
ewr: Unless there's a compelling reason to do this in MySQL, we should do this via the more Rails-standard:

```
  external_episodes.where("created_at < ?", self.days_to_expiry.days.ago)
```

ewr: I would also prefer to see this handled as a scope definition.

app/models/external_program.rb
```diff
@@ -125,6 +128,33 @@ def sync
...
+    end
+  end
+
+  def episodes
+    if days_to_expiry
+      external_episodes.where('DATE(?) < DATE_ADD(created_at, INTERVAL ? DAY)', Time.now, days_to_expiry)
+    else
+      external_episodes
+    end
+  end
```
ewr: I'm not convinced that we need to limit what episodes get viewed. Like I mentioned before, we should be conservative enough in our sweeping that all episodes still in the db are viewable, regardless of whether or not they are strictly past their expiration time.

Ravenstine:
You are right that nil can currently handle what MissingEpisode is currently doing.  Honestly, I was trying to be clever(as you have figured out).  Since we are only type checking once for the the @current_episode, it's really only replacing nil.  I can go in and change it to nil if that works better for us.  I do think that null objects(Missing) may be of use in the future, but probably not now opposed to what I initially thought.

Sure, we can remove the limitation on the episodes if the sweeper takes care of everything.  I included both options with the idea that we could either go with the sweeper or the query(or both) depending on what we'd decide.

I like your suggestion for the expired_episodes query.  I will also make it a scoped definition.

app/models/external_episode.rb
```diff
@@ -22,6 +22,13 @@ class ExternalEpisode < ActiveRecord::Base
     where("DATE(air_date) = DATE(?)", date_or_time)
   }
 
+  scope :expired, -> { joins(:external_program).where('CURRENT_DATE() >= DATE_ADD(external_episodes.created_at, INTERVAL days_to_expiry DAY)') }
```
ewr: I think this is cleaner as:

```ruby
scope :expired, -> { where('external_episodes.air_date < ?',self.external_program.days_to_expiry ? self.external_program.days_to_expiry.ago : 0) }
```

Less embedded SQL == more portable code. The ternary could also be replaced by a function on the program model, which returned days ago or 0 for models without expiration.

Also, I think `air_date` is the appropriate check here, not `created_at`.

app/models/external_program.rb
```diff
@@ -125,6 +124,23 @@ def sync
     self.importer.sync(self)
   end
 
+  alias_method :episodes, :external_episodes
```
ewr: Frankly, I don't understand why `episodes` isn't the relation. Why should external models use `external_episode`, `external_program`, etc?

I'd vote for standardizing the relationship names (instead of relying on an alias) unless it turns out that breaks something.

app/models/external_program.rb
```diff
@@ -125,6 +124,23 @@ def sync
...
+
+  def published_episodes
+    episodes.order("air_date desc")
+  end
+
+  def paginate_episodes episodes: nil, page: nil, per_page: nil, current_episode: nil
```
ewr: Are these functions still used?

Ravenstine: Nope.  Sorry for leaving in so much orphaned code.  Will be removing.

app/models/external_episode.rb
Ravenstine: I wanted to try something similar to that but can't get it to work because ```self```` is referencing the class in this context(so no self.external_program).  The only way other than using SQL in which I can see this working is by passing the external_program to the lambda as an argument, but that seems to kind of defeat the benefits of using a scope.

ewr: Huh. Interesting. 

In that case I would probably just abandon the scope and make `expired_episodes` a function on the program:

```ruby
  def expired_episodes
    if self.days_to_expiry
      self.episodes.where("air_date < ?",self.days_to_expiry.ago)
    else
      nil
    end
  end
```

Ravenstine: It's totally your call, but how about something like this:

```ruby
scope :expired, -> { joins(:program).where('air_date < ? - INTERVAL days_to_expiry DAY', Time.now) }
```
That way we aren't using any SQL functions and only the "INTERVAL" keyword, which appears to be supported in MySQL, PostgreSQL, Oracle SQL, and Amazon Redshift.

ewr: I believe that would be fine, but I still don't mind seeing this be entirely a function of the program and not the episode. I'm not going to argue strongly either way, though.

app/jobs/job/remove_external_episodes.rb
```diff
@@ -0,0 +1,15 @@
+module Job
+  class RemoveExternalEpisodes < Base
+    @priority = :mid
+
+    class << self
+      def perform
+        logger.info "Sweeping expired episodes form external programs. - #{Time.now}"
+        ExternalProgram.where('days_to_expiry IS NOT NULL').each do |program|
```
ewr: Let's make this a scope on ExternalProgram:

```ruby
scope :with_expiration, -> { where("days_to_expiry IS NOT NULL") }
```

ewr:
@Ravenstine: I believe the last step on this will be to reveal `days_to_expiry` in the Outpost admin GUI for external programs.

Ravenstine:
Ok, I added a form field for days_to_expiry and a validation.

ewr:
LGTM. When you come in this morning we can figure out a strategy for merge / deploy to staging, and then let's get this shipped.

PR #274, opened 2015-05-14
Fix episode preview in outpost by sharing behavior between the outpos…

Ravenstine:
Episode previews should now work in Outpost.


ewr:
Do you want to do individual pull requests for the episode redesign pieces, or just one for episode redesigns into master? I haven't looked into exactly what's entailed in the overall work.

sdillingham:
FWIW, I'd do these as commits within the episode-redesign branch, though I suppose there's no harm in parsing out the larger features into PRs.

PR #277, opened 2015-05-19
Episode redesign changes

Ravenstine:
- Episode preview
- Play button icon
- Option to hide comment count if count is zero
- Custom prompt message for play button(Listen to this Episode)


app/concerns/concern/controller/kpcc_episodes.rb
```diff
@@ -0,0 +1,21 @@
+module Concern
+  module Controller
+    module KpccEpisodes
```
sdillingham: Question for @Ravenstine and @ewr: do we want to name this module `KpccEpisodes` if it's responsible for rendering both KPCC episodes and external episodes?

sdillingham:
These changes look good to me @Ravenstine. At this point most of my concerns are about small design tweaks and polish that need to be applied before we deploy, but that can wait until this and the archive picker PRs are merged into `episode-redesign`. I'm happy to take the design/SCSS polish work once the functional pieces are all in place.

PR #278, opened 2015-05-19
Archive picker

Ravenstine:
I wouldn't merge this branch just yet; there are a couple of tests that fail which are fixed on another branch that I have been working on.  Those tests are unrelated to the archive browser.


sdillingham:
Great question @Ravenstine - yes, I think the smartest UI would be one where we only display the months and years we have content for on any given show.

PR #281, opened 2015-05-26
Add boolean to schedule_occurrences response to indicate if a pledge

Ravenstine:
…drive is happening.


Ravenstine:
I ended up going with an instance variable in the controller because I couldn't get a view test to work with the JSON builder, so it seemed like a better use of my time to write a controller test.

sdillingham:
This looks good to me. @ewr: any concerns to address before we merge and deploy?

ewr:
:+1: Merge and ship.

PR #282, opened 2015-05-29
Add Audio Player Events for GA

Ravenstine:


ewr:
General question: What's the value of putting the tracking on the PlayWidget, rather than on the Player itself? It seems safer to listen for events on jPlayer directly, rather than use click handlers.

Ravenstine:
That's a good idea.  I had put the tracking on the individual playwidgets because that way every piece of audio could have its own state, so that a user wouldn't repeatedly trigger quartile events if they replayed the same piece of audio.  Then again, maybe we don't need that?

app/assets/javascripts/audio.js.coffee
```diff
@@ -9,11 +9,42 @@ class scpr.Audio
...
+                time = e.jPlayer.status.currentTime
+                @states[@src()] ?= {}
+                @currentState = @states[@src()]
+                _.each [1,2,3], (i) =>
+                    if (time > (@duration() * i/4)) and !@currentState["quartile#{i}"]
+                        @currentState["quartile#{i}"] = true
+                        @sendEvent
+                            action: "Quartile#{i}"
+                            nonInteraction: true
+                            value: time
```
ewr: I don't know that time as a value helps us here. I would think value should just be 1.

app/assets/javascripts/audio.js.coffee
```diff
@@ -9,11 +9,42 @@ class scpr.Audio
...
             swfPath:  "/assets-flash"
             supplied: "mp3"
             wmode:    "window"
+            play: (e) =>
+                @states[@src()] ?= {}
+                @currentState = @states[@src()]
+                if @currentState.started != true
+                    @sendEvent
+                        action: 'start'
+                        nonInteraction: true
```
ewr: I think play should be an interaction event. @sdillingham: This is basically about bounce rates. If someone visits a page, plays audio, and then leaves, should they be considered a bounce? I tend to think no.

app/assets/javascripts/audio.js.coffee
```diff
@@ -9,11 +9,42 @@ class scpr.Audio
     constructor: (options={}) ->
         @options = _.defaults options, @DefaultOptions
 
+        @states = {}
```
ewr: I'm not sure I see the value in trying to track multiple files here. If a visitor plays a piece of audio, clicks to a different piece of audio, then clicks back and plays the first one again, I think that's actually two valid plays and should be reported as two events.

I would just track state for the current file and reset that when we get a `setmedia` event, unless there's something you're thinking of that I'm missing.

ewr:
Made a few comments, but I'm good with merging and deploying this, and then updating it as we go. The current events aren't giving us any utility, so there's no downside.

PR #284, opened 2015-06-02
Archive picker caching fix

Ravenstine:
Since we're already kicking @episodes, I figured I'd go ahead with the idea I had initially so I could move ahead with other things.  Let me know what direction we want to go in with this.

### DIFF
--- a/app/views/programs/kpcc/episode.html.erb
+++ b/app/views/programs/kpcc/episode.html.erb
@@ -145,36 +145,35 @@
 
   </div>
 </div><!--/ .report -->
-
+<% end %>
 
 
 
 <div class="episode-postlude">
   <div class="inner bound clearfix">
-
-
-    <section class="recents">
-      <h1>Recent Episodes from <%= @program.title %></h1>
-      <div class="articles clearfix">
-        <% @episodes.map(&:to_article).each do |episode| %>
-          <article>
-            <a href="<%= episode.public_path %>">
-              <h1><%= episode.short_title %></h1>
-              <div class="brief">
-                <% if episode.assets.present? %>
-                  <b class="asset"><%= render_asset episode, display: 'thumbnail' %></b>
-                <% end %>
-                <div class="synopsis">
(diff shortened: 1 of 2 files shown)


app/views/programs/kpcc/episode.html.erb
```diff
@@ -78,38 +78,35 @@
...
 
 
 
 <div class="episode-postlude">
   <div class="inner bound clearfix">
-
-
-    <section class="recents">
-      <header>
+    <% cache ["show", "episode-postlude-recents", "v1", @episode, @episodes.map(&:id)] do %>
```
ewr: Any time you're putting a cache key together, think about what changes would cause you to want to rebuild the cache.

In this case, why would you need to rebuild the cache if `@episode` changes?

Ravenstine: I guess not, especially if the recent episodes were shared between episodes.

app/views/programs/kpcc/episode.html.erb
```diff
@@ -78,38 +78,35 @@
...
-              <h1><%= episode.short_title %></h1>
-              <div class="brief">
-                <% if episode.assets.present? %>
-                  <b class="asset"><%= render_asset episode, display: 'thumbnail' %></b>
-                <% end %>
-                <div class="synopsis">
-                  <time datetime="<%= format_date(episode.public_datetime, with:'%F') %>"><mark><%= format_date(episode.public_datetime, with: '%A') %></mark>, <%= format_date(episode.public_datetime, with: '%B %e, %Y') %></time>
-                  <p><%= episode.teaser %></p>
+        <div class="articles clearfix">
+          <% @episodes.map(&:to_article).each do |episode| %>
```
ewr: Can you test and see if we actually need `to_article` here? We know we're only ever getting episodes, so I would rather this code not take the hit of casting to articles. I suspect a couple attribute names will need to change, but I think it should work otherwise.

PR #285, opened 2015-06-04
Use floats and round up in ContentBase pagination math.

### LINKED ISSUE #283: pagination UI not rendering final page number in pagination links
In any place where we're using the `paginate` helper to render pagination links, the final page of results doesn't get rendered in the list of page links. For example:

http://www.scpr.org/issues/the-fall-of-saigon-40-years-later

Shows only a single page of results, but if you pass in a second page to the URL:

http://www.scpr.org/issues/the-fall-of-saigon-40-years-later?page=2

You'll see two additional stories, so the paginate helper decided there was only one page of results when there were two. This is happening elsewhere:

Bio pages:

http://www.scpr.org/about/people/staff/maya-sugarman (displays two pages of results)
but a third page exists: http://www.scpr.org/about/people/staff/maya-sugarman?page=3

Ravenstine:
This should resolve #283.

### DIFF
--- a/app/models/content_base.rb
+++ b/app/models/content_base.rb
@@ -184,7 +184,7 @@ module ContentBase
       end
 
       define_method :total_pages do
-        ( @_pagination.total_results / @_pagination.per_page )
+        ( @_pagination.total_results.to_f / @_pagination.per_page.to_f ).ceil
       end
 
       define_method :offset_value do


ewr:
Silly, but seems like it would do the trick. :shipit:

sdillingham:
This looks good to merge and deploy @Ravenstine - go for it.

PR #288, opened 2015-06-08
Episode redesign changes

### LINKED ISSUE #268: Outstanding work on 'episode-redesign' branch
The `episode-redesign` branch has been paused WIP since January. With @Ravenstine on board, it's time to dust it off and :shipit:.

Markup and styling work was finished and QA'ed by @billofrights, so what's left is some Rails/front-end integration work, and a big shiny new front-end component - the Episode Archive Browser.

### Summary of tasks to complete:

- [x] FIX: Outpost episode preview broken (tests currently failing)
- [x] FIX: "Play button" triangle not rendering/animating in on episode audio
- [x] FEATURE: Implement "Originally Published" logic on published_at timestamps for segments
- [x] FEATURE: "Listen to this Story" should say "Listen to this Episode" on episode audio
- [x] FEATURE: Implement conditional Comment Count logic
- [x] FEATURE: Episode Archive Browser front-end component

### More Details

**Outpost episode preview**

Per my conversation with @Ravenstine, we'll want to implement the core episode template as a partial that can be used in both the public-facing view and Outpost-facing preview. We want to avoid some of the view duplication we've incurred in the recent past (e.g. `_segment.html.erb` and `_segment_preview.html.erb`).

**Play button triangle not animating in on episode audio**

This is a problem caused by how we scoped our javascript. The code to make this happen is here:

https://github.com/SCPR/SCPRv4/blob/episode-redesign/app/assets/javascripts/new/single.js#L46-L68

But single.js behaviors don't get loaded on episode views, and this code conditionally only applies the behavior to views that have a body class of "segment". We'll want to abstract this to be included in a more generic way.

**"Originally Published" logic**

Typically, all the segments in an episode's rundown will have the same published_at date, which is identical to the air date of the episode. Given that redundancy, we want to hide segment published_at dates by default, and conditionally display them only when the segment is significantly older than the episode (e.g. when we've pulled something out of our archives and included it in an episode. Let's set the logic to display "Originally Published MM/DD/YYYY" label when a segment's `published_at` is more than 4 weeks older than the episode's `air_date`.

**"Listen to this Episode" label**

  "Listen to this Story" label is hardcoded here:

https://github.com/SCPR/SCPRv4/blob/0cb59386bf6c669034954d570579eef64be31092/app/views/shared/cwidgets/new/_primary_audio.html.erb

We'll want to implement the ability to override that label in instances like this where "Listen to this Episode" is a more specific and appropriate label.

**Conditional comment count logic**

Currently this logic lives here:

https://github.com/SCPR/SCPRv4/blob/0cb59386bf6c669034954d570579eef64be31092/app/helpers/application_helper.rb#L386-L394

And is used in our old design templates to decorate article summaries with comment counts. For our new episode views, we need a similar helper that returns just the comment number when an article's comment count > 0, but allows us to hide the comment count and icon entirely when an article's comment count = 0.

Ravenstine:
This should resolve #268.


sdillingham:
It looks like the audio player buttons on old (pre-redesign) episode pages are breaking on this line of javascript:

https://github.com/SCPR/SCPRv4/blob/master/app/assets/javascripts/audio.js.coffee#L125

you can see an example by going to */programs/ask-me-another/ on dev or staging and trying to click any of the play buttons.

Issue #286, opened 2015-06-04
Some characters cause the preview for a news story to fail in Outpost.


Ravenstine:
For example:
http://www.scpr.org/outpost/news_stories/52185/edit

The apostrophe character(likely other UTF-8 characters too), when used in a headline or an asset caption, cause this error when the caption is rendered in a preview(from app/views/shared/assets/news/_photo.html.erb:13):

```Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8```

Here's the offending string from the headline: 

"California is sinking, and it’s getting worse"

... and the string from the asset caption:

"Michelle Sneed, a hydrologist with the geological survey, duplicated a photo by legendary hydrologist Joseph Poland to show a land sinking problem. Her photo, shown here, captures the early stages of today’s worsening subsidence problem."

The error occurs when the character is present in BOTH the headline and the asset caption.  It's possible that this applies to other text fields as well.  When the apostrophe in "today’s" is removed(or replaced with a tick), the error is not encountered and the preview renders.  The error also goes away if the apostrophe is removed from the headline.

This problem only seems to happen in production when previewing in Outpost.

https://rpm.newrelic.com/accounts/71496/applications/446356/traced_errors/63981c-c2238924-0ad8-11e5-88d6-c81f66b8ceca

ewr:
@Ravenstine: If you've verified this in production, go ahead and close it.

PR #292, opened 2015-06-15
Fix the outpost preview to reflect unsaved changes

Ravenstine:
My previous solution wasn't displaying updated news story content because, for some reason, Rails wasn't parsing nested attributes from the JSON form data.  I'd have preferred that solution but I couldn't get it to work without having to do something kind of janky on the SCPRv4 side.

So instead I have reverted to using the version of Outpost before my previous change and am now transforming the keys of the news story attributes from the parameters.  I've previewed dozens of news stories in staging and they all seem to work and reflect changes that have not yet been saved.

If anyone has any suggestions on how to better accomplish this, please let me know.  I made a bunch of different attempts that ended up having different side-effects.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ gem 'render_anywhere'
 gem 'dalli'
 
 ## Outpost
-gem 'outpost-cms', path: "~/workspace/outpost"
+gem 'outpost-cms', github:"SCPR/outpost", tag:"v0.1.6"
 gem 'outpost-publishing'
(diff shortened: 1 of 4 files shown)


ewr:
it would be nice if this also got a test for the data that was getting treated as binary before the work on #286 began. 

That said, it's important to fix preview quickly, either by pushing this or reverting the work that was previously merged. Do whichever you feel most confident in.

PR #295, opened 2015-06-16
Remove float value that was being sent with audio complete event.

Ravenstine:
Even though I can't immediately confirm this with my GA account(Real-time is not reflecting all of the events that I'm sending it), I have a hunch that most of the 'complete' events were not showing up because GA expects an integer as the event value and not a float; the event was still sending a value, which I had initially set to the current time of the audio.  If we look at the 'complete' events that do show up in GA, they all have non-float values(which is likely to happen occasionally).

Instead, I have changed it to just send a value of 1.  I'd like to merge this and wait and see if these events start coming in.


ewr:
We might as well push this and just see if it works.

Issue #279, opened 2015-05-21
URL Validator should check for trailing empty spaces


sdillingham:
Currently, our URL validator for ActiveModel (https://github.com/SCPR/SCPRv4/blob/master/lib/validators/url_validator.rb) doesn't throw an error when the URL string is followed by trailing spaces. The presence of trailing spaces can break functionality in our views, so we should extend the functionality of our URL validator to defend against this.

PR #296, opened 2015-06-17
Add sanitization concern to clean up trailing whitespaces in model URL attributes.

Ravenstine:
Addresses issue #279.  Adds a method that can be called by a 'before' callback which will strip preceding and trailing whitespaces in specified URL attributes.


ewr:
:+1: Looks sensible to me.

sdillingham:
Hey @Ravenstine, I just got an email from the newsroom about trailing whitespace on URLs happening in the Short List. Looks like we need to add the concern to this model as well:

https://github.com/SCPR/SCPRv4/blob/master/app/models/abstract.rb

Could you implement that when you have a moment?

Issue #240, opened 2015-03-03
NPR Ingests Need to Adapt After Failures


ewr:
On Thursday morning, scheduled tasks were not run for the first six to eight hours of the day. When scheduled tasks restarted, the NPR importer started picking up new content, but did not pick up most content that had been published during the outage.

The NPR importer is hardcoded to use the last hour as its search:

https://github.com/SCPR/SCPRv4/blob/4b19db4168b95ae4d0693f9f0831254620e38c02/app/importers/npr_article_importer.rb#L26

That should be changed to be something like the earlier of the last hour or the last NPR article imported.

PR #297, opened 2015-06-18
Change the NPR article importer to sync from the last RemoteArticle p…

Ravenstine:
Addresses issue #240.  I have the importer looking for NPR stories since the last NPR Remote Article downloaded.


ewr:
Wouldn't this fail to get all stories if more than 20 stories were published since the last import?

Issue #263, opened 2015-05-07
Non-UTF8 character in filename causes file uploads to fail


ewr:
After the Rails 4.2 upgrade, if an audio file is uploaded with a funky filename, it will raise an exception.

The actual failing code is in `action_dispatch/http/upload.rb:31`:

```ruby
    @original_filename &&= @original_filename.encode "UTF-8"
```

We either need this to be fixed in Rails, or we need to clean the filename before it gets to this point.

Ravenstine:
Is it Encoding::UndefinedConversionError?  I'd think that Rack-utf8_sanitizer would take care of that, but maybe it doesn't force encoding by removing invalid bytes.

ewr:
Here’s a New Relic error report:

https://rpm.newrelic.com/accounts/71496/applications/446356/traced_errors/3355555435

Feel free to investigate. :)

From: Ravenstine
Reply-To: SCPR/SCPRv4
Date: Friday, May 8, 2015 at 12:08 PM
To: SCPR/SCPRv4
Cc: Eric Richardson
Subject: Re: [SCPRv4] Non-UTF8 character in filename causes file uploads to fail (#263)


Is it Encoding::UndefinedConversionError? I'd think that Rack-utf8_sanitizer would take care of that, but maybe it doesn't force encoding by removing invalid bytes.

Ravenstine:
I haven't found any existing solution that does what we need.  I've also tried utf8-cleaner.  My guess is that if any of these utf8 sanitizing gems work, they aren't going deep into the response object; since our file upload is nested inside of form params, they may not even find the file name.

I hacked up this Rack middleware that finds nested hash keys called "filename" and removes invalid byte characters from their string values.  I'm not sure if this how we'd want to go about it, but it's my only proof of concept for how we could fix file names without monkeypatching ActionDispatch.

```ruby
class Sanitizer
  def initialize app
    @app = app
  end
  def call env
    env["rack.request.form_hash"] = sanitize(env["rack.request.form_hash"])
    @app.call env
  end
  def sanitize input
    if input.respond_to? :each_key
      input.each_pair do |k, v|
        input[k] = sanitize_pair k, v
      end
    end
    input
  end
  private
  def sanitize_pair key, value
    if key.to_s.match(/file[_\s]*name/i) && value.is_a?(String)
      value.encode "UTF-8", invalid: :replace, undef: :replace, replace: "_"
    elsif value.is_a? Hash
      sanitize value
    else
      value
    end
[...]

Issue #248, opened 2015-03-30
Feature audio id needs to strip / throw error on invalid characters


ewr:
Just saw this issue in New Relic:

`URI::InvalidURIError: bad URI(is not URI?): http://media.scpr.org/audio/features/20150326_features1076 .mp3`

The issue is the space between `1076` and `.mp3`.

We need a validation on `Audio.enco_number` to make sure whitespace (or any other non number characters) can't get inserted.

Ravenstine:
I actually worked on this a while ago but didn't make a pull request for it.  Had to change it slightly, though.  

https://github.com/SCPR/SCPRv4/pull/321

PR #300, opened 2015-06-22
File name sanitize

Ravenstine:
Addresses issues #263.

If a "safe_filename" input is included inside a form, the SafeFilename middleware finds any key in the form that matches "filename" and with the value replaces whitespaces with underscores and accented characters with their simple equivalents.


ewr:
Why do we care to be able to turn this on/off in the form? Wouldn't turning it off still just cause it to fail to upload?

Ravenstine:
I added the ability to toggle it so that we aren't recursively searching a form hash every time we submit a form that doesn't have file names or uploads of any kind.  Not that it's that big a deal since it's only on submits and the form hash is never very deep.  If we don't care about this, then we don't need the on/off aspect.

ewr:
I don't believe this does anything for #248. That's a form validation issue, not an upload issue.

I believe we would just want this applied to any uploads. I'm not a fan of that being something the form itself deals with, but potentially it could be via specifying a path in a config file or something like that if you think it's necessary for performance.

Ravenstine:
I mistakenly believed that issue had to do with the file upload.

I would not mind omitting the switch all-together if we want this functionality to apply to all(or most) of our forms; if there ever comes a time where we don't want the file name to be sanitized, perhaps it would be better to specify a blacklist of paths in a config file.  Performance would only be an issue when someone submits a form, and any difference in performance would probably be imperceptible.  Sometimes I'm overly-cautious.

PR #301, opened 2015-06-23
Prevent news stories from crashing on save when being published and having tags added

Ravenstine:
There was an issue with news stories trying to update tags with its published_at timestamp before it had a chance to assign its published_at.  This fixes that by checking for the published_at existence before trying to update the tags and then let the tags get updated on after_save when the published_at has now been added.


ewr:
:shipit:

PR #302, opened 2015-06-23
Make NPR article importer adapt to failures and download more than 20 articles at a time.

Ravenstine:
#240

This extends upon my original change, except the importer will now download more than 20 articles by downloading multiple batches of 20(which is a limit of the NPR API).  I have confirmed that it downloads more than 20 unique articles at a time.

### DIFF
--- a/app/importers/npr_article_importer.rb
+++ b/app/importers/npr_article_importer.rb
@@ -23,10 +23,11 @@ module NprArticleImporter
       # more often than that!
 
       npr_stories = []
-      offset    = 0
-
+      offset      = 0
+      start_date  = RemoteArticle.where(source: "npr").last.try(:published_at) || 1.hour.ago
+      end_date    = Time.zone.now
       begin
-        response  = fetch_stories(offset)
+        response  = fetch_stories(offset, start_date, end_date)
         npr_stories   += response
         offset    += 20
       end until response.size < 20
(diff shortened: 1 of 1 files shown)


app/importers/npr_article_importer.rb
```diff
@@ -59,6 +61,15 @@ def sync
       added
     end
 
+    def fetch_stories offset
+      NPR::Story.where(
+          :id     => IMPORT_IDS,
+          :date   => ((RemoteArticle.where(source: "npr").last.try(:published_at) || 1.hour.ago)..Time.zone.now))
```
ewr: This timestamp should be set once at the start of the run and passed into fetch_stories.

app/importers/npr_article_importer.rb
```diff
@@ -59,6 +61,15 @@ def sync
...
+        .set(
+          :requiredAssets   => 'text',
+          :action           => "or")
+        .order("date descending").limit(20).offset(offset).to_a
```
ewr: I wonder if this sort should be ascending... Since we're filling in data, we wouldn't want to crash part way but have masked the gap for later by creating new content.

I can't think of anything that ascending would mess up.

Ravenstine: Sounds good to me.

PR #303, opened 2015-06-24
Add form hash filter to force the encoding of binary form strings to UTF-8

Ravenstine:
@ewr @sdillingham 

I have now changed the implementation to be included in the same place that the file name sanitization happens.  A middleware traverses the Rack form hash and calls #force_encoding on every string so that binary strings are now interpreted as UTF-8.  If for some reason force_encoding fails, it will just return the original string untouched.  

So far, I think this is the best approach because since binary files uploaded through the form have already been saved to tempfiles and referenced by File objects, they'd be ignored by my code since they are not strings.  If we are already having our files uploaded in that fashion, I can't think of any reason why we would ever need binary to come through in our form input.  

Combining this with our file name sanitization hopefully means that we wouldn't have as many encoding problems.


ewr:
Should this be closed now that we're tackling this via removing the monkeypatch?

Ravenstine:
I think that depends on how long we want to wait to see if there are any issues with the spider crawls(without the monkeypatch).  If we don't want to wait that long, then I don't see why this would need to remain open.

Ravenstine:
Actually, I'll close this since our solution has already been merged.  If that ends up being problematic, this can be re-created.

PR #308, opened 2015-06-30
Add content sanitizer and make the sanitizer concerns more uniform.

Ravenstine:
@ewr @sdillingham 

Look good?

### DIFF
--- a/app/concerns/concern/sanitizers/content.rb
+++ b/app/concerns/concern/sanitizers/content.rb
@@ -6,8 +6,11 @@ module Concern::Sanitizers::Content
   end
 
   def remove_bad_characters
-    [:name, :title, :short_title, :headline, :slug, :teaser, :body, :byline].each do |attrib|
-      try(attrib).try(:gsub!, Regexp.new("\u2028|\u2029"), "")
+    attribute_names.each do |k|
+      v = send(k)
+      if v.is_a?(String)
+        v.gsub!(Regexp.new("\u2028|\u2029"), "")
+      end
     end
   end
 end
--- a/app/concerns/concern/sanitizers/content.rb
+++ b/app/concerns/concern/sanitizers/content.rb
@@ -9,7 +9,7 @@ module Concern::Sanitizers::Content
     attribute_names.each do |k|
       v = send(k)
       if v.is_a?(String)
-        v.gsub!(Regexp.new("\u2028|\u2029"), "")
+        send "#{k}=", v.gsub(Regexp.new("\u2028|\u2029"), "")
       end
     end
   end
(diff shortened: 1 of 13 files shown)


app/concerns/concern/sanitizers/content.rb
```diff
@@ -0,0 +1,11 @@
+module Concern::Sanitizers::Content
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :remove_bad_characters
+  end
+
+  def remove_bad_characters
+    body.gsub!(Regexp.new("\u2028|\u2029"), "")
```
ewr: The characters can appear in more than just `body`. The most recent example was an asset caption, but it could also be teasers, headlines, etc.

Ravenstine: Would something like this be comprehensive enough?  

```ruby
    [:name, :title, :short_title, :headline, :slug, :teaser, :body, :byline].each do |attrib|
      send(attrib).try(:gsub!, Regexp.new("\u2028|\u2029"), "")
    end
```

I got those from looking through most of the models that display content.  

Or do we want to apply this sanitizer to every string on a model?

ewr: I'm generally wary of just throwing everything at it and seeing what sticks via `try`. I can't think of a good argument against applying the gsub to any string attribute, though. You could iterate `self.attributes` or something along those lines.

app/concerns/concern/sanitizers/content.rb
```diff
@@ -0,0 +1,16 @@
+module Concern::Sanitizers::Content
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :remove_bad_characters
+  end
+
+  def remove_bad_characters
+    attribute_names.each do |k|
```
Ravenstine: I ended up using #attribute_names because #attributes returns attribute values from before modifications, so body would always come up as nil even if I modified it through an attribute.  This way I can check attributes that have changed but have not yet been saved.

app/concerns/concern/sanitizers/content.rb
```diff
@@ -0,0 +1,16 @@
...
+      v = send(k)
+      if v.is_a?(String)
+        v.gsub!(Regexp.new("\u2028|\u2029"), "")
```
ewr: My only last thought is to wonder if this should be something like: 

`send("#{k}=",v.gsub(Regexp.new("\u2028|\u2029"), "")`

just to account for the idea that you might overwrite an attribute setter in a model.

I don't think that would have any practical impact on our code right now, though.

Ravenstine: That makes a lot of sense, actually.  Will make that change and deploy.

PR #310, opened 2015-07-09
Schedules

Ravenstine:
Just opening this for the time being in case anyone wants to glance at the code.

What I've done is the following:

- Reduce the "lookahead" to 2 weeks from 2 months.
- Job runs once per day.
- When a rule is changed, future occurrences are destroyed and replaced with the new ones.  No checking for duplicates.


ewr:
So every time any rule is changed you're rebuilding the entire set of future schedule occurrences? Why is that better than just rebuilding occurrences for the rule being updated?

Ravenstine:
The job rebuilds future schedule occurrences for all rules.  When a rule is updated, the future occurrences for that rule are rebuilt but the future occurrences for other rules remain intact.  Non-recurring schedule occurrences are not effected.

I think this is better because, instead of finding "duplicates" based on existing occurrences with the same exact same start time, this doesn't leave as much room for overlapping or other schedule conflicts.  The only exception is non-recurring occurrences, which are created through user-intervention so I'm not sure we should worry about them.

app/models/recurring_schedule_rule.rb
```diff
@@ -138,80 +155,59 @@ def build_schedule
...
+        soft_starts_at: occurrence.start_time + (soft_start_offset || 0),
+        program:        program,
       )
     end
+    schedule_occurrences
+  end
 
-    self.schedule_occurrences
+  def rebuild_occurrences(args={})
+    execute_then_destroy_old_occurrences do
```
ewr: If this is just building, but not saving, shouldn't the destroy wait until the save rather than here?

Ravenstine:
I'm going to leave this in staging for the time-being to observe how it's building articles.  I'll the job once per morning and check to see if it's building the expected schedule.

Issue #312, opened 2015-07-23
Implement newsletter signup on story pages


vsinghkpcc:
AC

1) There is a way to signup for the daily short list / breaking news newsletters on all story pages
2) We track as close as possible to: 
- "how many users saw this" 
- "Of the users that saw this, how many subscribed"
- "Was there a specific article that had a high conversion rate to subscription"

PR #318, opened 2015-08-04
Newsletter signup

sdillingham:
* Renames existing newsletter subscribe partial to scope it only to breaking news
* Creates a new partial for the Short List subscription form
* Adds a new SCSS file for styling this module
* Adds the partial to our news, blog and segment article views


sdillingham:
Addresses #312

PR #321, opened 2015-08-07
Enco id validation

Ravenstine:
#248

Adds a validation so that the enco_number can only be numeric, so this rejects word characters and whitespace.


Ravenstine:
I've got to fix some tests first.

PR #347, opened 2015-09-29
Inline assets

Ravenstine:
Some of these commit messages need to be renamed; I attempted to squash some of my commits but haven't done it to all of them.  For review purposes only.

Related review PRs:
https://github.com/SCPR/outpost/pull/7
https://github.com/SCPR/outpost-asset_host/pull/1


app/assets/stylesheets/new/_newsletter_appeals.css.scss
```diff
@@ -25,26 +25,6 @@
     }
 }
 
-// This element will be used to create a dark overlay
```
ewr: Not part of inline assets?

app/assets/stylesheets/new/single.css.scss
```diff
@@ -634,6 +634,38 @@ body.single .comments {
         }
 
 
+        // "Part of a Series" citation:
```
ewr: Not part of inline assets?

app/views/podcasts/index.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="row-fluid">
   <div class="span6">
-    <p style="font-style: italic">A podcast is a way for you to "subscribe" to a program and listen to it when and where you want. It's kind of like TiVo for radio, with the added benefit of portability.</p>
```
ewr: Not part of inline assets?

app/views/shared/header/_document_header.html.erb
```diff
@@ -36,6 +36,19 @@
     }
   </script>
   <script type="text/javascript" src="//nexus.ensighten.com/mogointeractive/1078/Bootstrap.js"></script>
+
```
ewr: Not part of changes for inline assets?

public/static/ckeditor/build-config.js
```diff
@@ -66,7 +66,7 @@ var CKBUILDER_CONFIG = {
 		'tabletools' : 1,
 		'toolbar' : 1,
 		'undo' : 1,
-		'wysiwygarea' : 1
+		'wysiwygarea' : 1,
```
ewr: Was this intentional? This traditionally causes an exception in IE.

lita/lita-scprv4/Gemfile.lock
```diff
@@ -0,0 +1,90 @@
+GIT
```
ewr: Not part of inline assets?

ewr:
`InlineAssets.render` was doing some things it didn't need to, and I think life is going to be better in the long-run if we standardize with the existing flow of the `render_asset` helper. That allows us to continue to have one spot for template selection hierarchy, etc.

I did throw out the code related to asset "shape". If that is something that we want to use in our styling, we should look at where best to implement it. Assethost already computes asset shape, so we could potentially just expose that through the JSON outputs:

https://github.com/SCPR/AssetHost/blob/e786cebc31f51e0b26f02b023a8a8010b69792c1/app/models/asset_host_core/asset.rb#L126-L151

app/views/podcasts/index.html.erb
Ravenstine: I'm not certain why that change @sdillingham made is not on master, but this was the commit:
https://github.com/SCPR/SCPRv4/commit/f4ec6ffd95a81e5a826feb8efaa7625defd3f41b

public/static/ckeditor/build-config.js
Ravenstine: Intentional, but didn't know that caused an exception in IE.  Will remove.

app/helpers/application_helper.rb
```diff
@@ -157,12 +159,38 @@ def render_asset(content, options={})
...
+      # assets. in any case where we're rendering a body, we'll already have
+      # the original object loaded, so that's ok
+      asset = content.original_object.assets.find_by(asset_id:asset_id)
+
+      if asset
+        rendered_asset = render_asset content, context:"news", display:"inline", asset:asset
+        placeholder.replace Nokogiri::HTML::DocumentFragment.parse(rendered_asset)
+      else
+        # FIXME: I'm sure there's a cleaner "delete"
+        placeholder.replace Nokogiri::HTML::DocumentFragment.parse("")
```
Ravenstine: @sdillingham @ewr Do we want to delete the placeholder instead of leaving it there?  I'm cool either way, but it seemed like some of the decision making behind using an img tag as the placeholder was so that it would serve as the fallback if we don't get a good response from AssetHost.

ewr: We're rendering a view for end users. I can't think of a reason an empty placeholder is a better UI for them than seeing nothing.

We're also not typically querying Assethost here. The actual asset should be in cache.

Issue #345, opened 2015-09-25
API should only attach `via=api` to our own audio


ewr:
SCPRv4 applies a `via=api` to audio links in the episode API output. This is only useful for SCPR audio, and shouldn't be applied to external content.

PR #359, opened 2015-10-26
Only show via=api param for our own episodes in the audio API.

Ravenstine:
#345 

The url_with_params method has been changed to add the "via=api" param to links that match "scpr.org".


app/views/api/public/v3/audio/show.json.jbuilder
```diff
@@ -1,5 +1,5 @@
 json.partial! api_view_path("shared", "meta")
 
 json.audio do
-  json.partial! api_view_path("audio", "audio"), audio: @audio
+  json.partial! api_view_path("audio", "audio"), audio: @audio, content: @content
```
ewr: What is this change doing?

ewr:
Audio can also come from show segments, so those need to be addressed as well.

ewr:
Actually, audio can also come in via a NewsStory, and probably a BlogEntry... I think it's going to be more efficient to scrap this approach and look at making the change in `url_with_params` based on looking at the URL there.

Ravenstine:
I'm not sure I'm sold on that logic belonging in `url_with_params`, unless we have something different in mind.  It would be simple enough to just give these models a uniform way to tell us whether or not they are KPCC.  While it would be easy to check the name of the class in somewhere like `url_with_params`, some models like NewsStory would require a different approach because it actually stores its source as a field, and there are several different source names.  To me it sounds like those are logic that should be model-specific.

app/views/api/public/v3/audio/show.json.jbuilder
Ravenstine: @ewr Actually, it's a semi-redundant fix for something that I actually don't fully understand.

In _collection.json.jbuilder, there's some code at the top defaulting ```content``` to nil in case @content isn't being set by the controller, since it's a variable being expected by _audio.json.jbuilder.  What I find strange is that @content doesn't seem to be set in any controller that is rendering audio JSON.  I'm guessing that it's just always deriving the context from the content related to the audio and not content that is being passed in as a local variable.

So I guess a question would be if we even need this local variable(content) at all and, if not, can we just remove it?  Or maybe we can keep the option there and just change _audio.json.jbuilder to not fail on the absence of the variable.

Issue #365, opened 2015-10-28
News Audio Gets No Via or Context Params


ewr:
The code to render audio into news story gives them no params, so we don't get any reliable analytics for those plays on the backend: 

![screen shot 2015-10-28 at 4 17 14 pm](https://cloud.githubusercontent.com/assets/204350/10801769/6927a95e-7d8f-11e5-972e-b1db37a0f438.png)

Audio on news stories should be tagged with `via=website&context=news`.

Issue #358, opened 2015-10-23
Website Audio Doesn't Always Include Show Context


ewr:
We use a `context=<SHOW>` query param to connect audio downloads with shows. More than 50% of our non-podcast audio requests currently come in missing this parameter, and it seems that's at least partially because we aren't including it all the time when we render audio players.

All first-party audio that we render on the site needs to include a valid context (show or "news", primarily) so that we can have working audio analytics.

Taking a quick look at the code I see several templates that are calling the "primary_audio" or "extra_audio" content widgets without a context option, but there may be other audio calls that I'm missing.

PR #368, opened 2015-10-30
Provide context in audio links that don't currently have it

Ravenstine:
#365 #358

Audio links that did not include context & via params now do.  Will add more if any are found.

### DIFF
--- a/app/views/shared/new/_single_blog.html.erb
+++ b/app/views/shared/new/_single_blog.html.erb
@@ -13,7 +13,7 @@
                         <% audio.each_with_index do |audio, index| %>
                             <li class="story-audio">
                             <li class="story-audio" data-ga-category="Article" data-ga-action="Play" data-ga-label="Player">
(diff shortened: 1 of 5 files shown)


app/views/shared/new/_single_blog.html.erb
```diff
@@ -13,7 +13,7 @@
                         <% audio.each_with_index do |audio, index| %>
                             <li class="story-audio">
                             <li class="story-audio" data-ga-category="Article" data-ga-action="Play" data-ga-label="Player">
-                                <a href="<%= audio.url %>" class="play-btn audio-toggler" title="<%= h(article.short_title) %>" data-duration="<%= audio.duration %>">
+                                <a href="<%= raw url_with_params audio.url, via: 'website', context: @episode.show.slug %>" class="play-btn audio-toggler" title="<%= h(article.short_title) %>" data-duration="<%= audio.duration %>">
```
ewr: Episode? This is the blog template. I would assume `article.blog.context` or something like that?

Issue #361, opened 2015-10-27
Turn KPCC Plus web player off outside of drives


vsinghkpcc:
AC

- [ ] KPCC Plus web player is "turned off" in a way that doesn't break the experience for a user.

vsinghkpcc:
Notes:
- Vijay will provide copy for messaging
- Use alternate template to message to user "Drive over, KPCC Plus no longer available"

vsinghkpcc:
#Sorry, KPCC Plus is Off-Air

##We aren’t in Membership Drive mode right now.

Not to worry - simply [visit the Listen Live page] to hear the normal programming you know and love.

If you are a current member, we’ll e-mail you a link to access KPCC Plus again during our next drive. For more information about KPCC Plus, please visit the FAQs page.

Happy listening!

Ravenstine:
The pledge free stream has been turned off and replaced with a message page.  Error page is no longer a Flatpage.

PR #372, opened 2015-11-03
KPCC Plus: Display alternative message to members when a pledge drive is over

Ravenstine:
#361
The pledge free stream now redirects to a flatpage with a message and a link to the regular live stream.

image


Ravenstine:
There's a problem with the migration.  I'll repost this.

sdillingham:
@Ravenstine, we had some back and forth in sprint planning about whether the "off air" page should be a flatpage or not, but in the end we landed on the idea of it being a view managed in the code rather than a flatpage managed in the DB.

Could you revise the PR to add the "off-air" messaging as separate view; maybe something that lives at `app/views/listen/pledge_free-off_air.html.erb`? If the view inherits the application.html.erb layout it should look the same as it would if it were a flatpage, but be easier to manage in code.

sdillingham:
While you're at it, it would be great if you could take the flatpage content that currently lives at http://www.scpr.org/pledge-free/error and move it into its own view as well, just so we're being consistent about keeping all these KPCC Plus notification/error messages in code.

PR #373, opened 2015-11-03
Redirect members from KPCC Plus to off air page

Ravenstine:
#361
*It won't let me reopen the other pull request because I force pushed.*
The pledge free stream now redirects to a flatpage with a message and a link to the regular live stream.

Instead of using Flatpages, I changed the error page and off air page to be their own templates.

The only thing that's different than what is represented in the image below are the share icons, which don't share up with this implementation.

EDIT: Oh, and I linked the FAQ page reference to spcr.org/pledge-free.

image


sdillingham:
Looks good to me, @Ravenstine! I'd say :shipit:.

Issue #366, opened 2015-10-29
Remove Social Sharing Numbers on Articles


vsinghkpcc:
AC

- [x] Facebook and Twitter sharing numbers no longer display on articles
- [x] Clean up any associated code across scprV4

Ravenstine:
The social counts have been removed.

PR #375, opened 2015-11-04
Remove social counts for Facebook and Twitter.

Ravenstine:
#366

I've removed the JavaScript and HTML relating to Facebook and Twitter social counts.

### DIFF
--- a/app/assets/javascripts/utilities.js.coffee
+++ b/app/assets/javascripts/utilities.js.coffee
@@ -120,8 +120,6 @@ class scpr.SocialTools
         disqfinder      : ".social_disq"
         count           : ".count"
         gaq             : "_gaq"
-        fburl           : "http://graph.facebook.com/fql"
-        twiturl         : "http://urls.api.twitter.com/1/urls/count.json"
         disqurl         : "https://disqus.com/api/3.0/threads/set.jsonp"
         disq_api_key    : 'EAlRB1x2Coj262YxEGbmvecZufapzC8IJ8aqKgtG0ILS4L1EMucA66Jq9MlY270b'
         disq_short_name : 'kpcc'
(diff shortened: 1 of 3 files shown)


sdillingham:
Functionally this looks good @Ravenstine, and I think you caught all the templates where we had markup to remove. A couple minor edits:

The `twiturl` option in the scpr.SocialTools class is no longer used, and can be removed:

https://github.com/SCPR/SCPRv4/pull/375/files?diff=unified#diff-f69e34f5df6c22bc6b3c7aa39f45fc7bR124

Originally, we were setting the `fburl` option to default to Facebooks FQL endpoint:

https://github.com/SCPR/SCPRv4/pull/375/files?diff=unified#diff-f69e34f5df6c22bc6b3c7aa39f45fc7bL123

But since we're not using that endpoint anymore to get FB share counts, that default doesn't make sense. Now `fburl` is only being set to wire up the sharing functionality:

https://github.com/SCPR/SCPRv4/pull/375/files?diff=unified#diff-f69e34f5df6c22bc6b3c7aa39f45fc7bL166

Maybe there's a more sensible default for `fburl` given the way we're using it?

It's a small detail, but this comment should be updated to remove the reference to "fetch(ing) share counts" since we're not doing that anymore:
  
https://github.com/SCPR/SCPRv4/pull/375/files?diff=unified#diff-f69e34f5df6c22bc6b3c7aa39f45fc7bL134

Ravenstine:
@sdillingham Taking a further look, I can't come up with a more sensible value for `fburl` variable since the other uses of facebook.com urls require a specific scope.  Thanks for catching those details that need to be changed.

sdillingham:
:+1:

sdillingham:
Looks good @Ravenstine. This is ready to :shipit:.

Issue #378, opened 2015-11-10
Shortlist Newsletter (Daily and Monday) Ad Placement


vsinghkpcc:
AC
- [ ] The ad is placed after the 4th story in the Shortlist Newsletter (Daily and Monday)

sdillingham:
The ad block currently sites here:

https://github.com/SCPR/SCPRv4/blob/master/app/views/editions/email/template.html.erb#L167-L175

And an empty ad slot block exists in the Monday Short List template here:

https://github.com/SCPR/SCPRv4/blob/master/app/views/editions/email/template.html.erb#L167-L175

It should be replaced with the block from the first file.

Ravenstine:
Here's a screenshot of what happens when I place the ad between a slice of the first 4 headlines and the remaining headlines.  Looks the same for both the regular and Monday Short List.

Is this what we are expecting?
![short-list-ad](https://cloud.githubusercontent.com/assets/5193330/11132918/9045343e-8947-11e5-8c7e-624075957df4.png)

sdillingham:
That's close... let's try using this block of markup for the ad instead:

https://gist.github.com/sdillingham/355ae361506789638c94

It should give us a horizontal rule above the ad, and center the ad in the middle of the column.

Ravenstine:
@sdillingham That does look better:
![image](https://cloud.githubusercontent.com/assets/5193330/11133351/36d3320e-894a-11e5-981a-3071ec19a5f9.png)

sdillingham:
Yep, that's the jam. Looks good!

PR #380, opened 2015-11-12
Short List Ad Placement

Ravenstine:
#378

An ad banner is now placed after the 4th(at most) story in both version of the Short List.

image


sdillingham:
Reviewed this and it looks good @Ravenstine. :shipit:

Issue #384, opened 2015-11-16
Short List Email Sends Need to be More Resilient


ewr:
Three times in the last week short list emails have failed to send, twice due to hitting a 500 when POSTing to Eloqua. 

In each case I've manually gone into the console to find the appropriate EloquaEmail object and then manually triggered `async_send_email`.

Our send process needs to automatically retry email sends that fail, particularly for an issue like a 500 on Eloqua's side.

Ravenstine:
I know this is happening with EloquaEmails but, out of curiosity, has it been happening with BreakingNewsAlerts?

sdillingham:
No reports of this happening on the Breaking News Alerts, but obviously it would be smart to set any eloqua send to retry if it fails.

sdillingham:
How much of a delay were you thinking of introducing? I wouldn't wait more than 60 seconds to retry.

Ravenstine:
Yeah, I wasn't thinking much more than that.  I think 60 seconds and 3 total tries makes sense.

Ravenstine:
@sdillingham @ewr 

@ewr and I chatted the other day and landed on the idea of using a job to resend emails, so that's the direction I've now gone in.

So far, I have two versions of changes.  Both are very similar, but there are some differences.  Some peer review would be appreciated.

## Version 1
### Comparison
https://github.com/SCPR/SCPRv4/compare/eloqua_retry_spike3
### Summary
A job called `ResendEmails` Looks for EloquaEmails that have already been attempted to send at least once but failed to send, then tries to resend them by calling `#publish_email`.  Note that I haven't actually scheduled the job to run with Cron because doing so is trivial and I just haven't gotten to it.  I constrained the Job to only act upon EloquaEmails created after 11/17/2015 because I'm paranoid.

## Version 2
### Comparison
https://github.com/SCPR/SCPRv4/compare/eloqua_retry_spike4
### Summary
Almost the same concept as seen in Version 1, but the `EloquaSendable` concern module has been removed and the functionality moved into the `EloquaEmail` model.  Anything that sends an email now does so by creating EloquaEmails, now including `BreakingNewsAlert`.  Emails are no longer queued up with Resque; instead, the `SendEmails` job attempts to send all emails that haven't been sent yet that haven't been attempted more than 3 times.  Test currently fail under this branch(and some need to be moved into eloqua_email_spec), but I'll of course address this.

## Question
[...]

PR #386, opened 2015-11-20
Merge EloquaSendable into EloquaEmail and remove the concern module

Ravenstine:
#384 

As described in the issue:

The EloquaSendable concern module has been removed and the functionality moved into the EloquaEmail model. Anything that sends an email now does so by creating EloquaEmails, also now including BreakingNewsAlert. Emails are no longer queued up with Resque; instead, the SendEmails job attempts to send all emails that haven't been sent yet that haven't been attempted more than 3 times.

## Notes

The `send_at` concept is going away, and was mostly a relic of my first pass.  As @ewr mentioned, we still need to work on the concept of status messaging for BreakingNewsAlerts since we are changing how it is sending emails and need to make sure that it either works the same or is improved.  Of course, I would prefer the latter if possible.

The way that a 500 status is determined is problematic so any thoughts on how that can be better re-worked are welcome.  

Looking at our log this morning, our problem is being caused by a `Faraday::Error::ConnectionFailed` exception.


app/jobs/job/send_emails.rb
```diff
@@ -0,0 +1,17 @@
+module Job
+  class SendEmails < Base
+    @priority = :mid
+
+    class << self
+      def perform
+        logger.info "Sending unsent emails to Eloqua. - #{Time.zone.now}"
+        emails = EloquaEmail.where(email_sent: false)
```
ewr: All of this logic should be a scope in EloquaEmail, so that here it's just `EloquaEmail.unsent.each`...

app/jobs/job/send_emails.rb
```diff
@@ -0,0 +1,17 @@
...
+          .where(attempts_made: 0..2)
+          .where("send_at <= ? OR send_at IS NULL", Time.zone.now)
+          .where("created_at > ?", Date.parse("2015-11-17"))
```
ewr: Rather than an embedded date like this, I'd prefer a migration to destroy any outdated unsent emails.

app/models/breaking_news_alert.rb
```diff
@@ -59,6 +58,7 @@ class BreakingNewsAlert < ActiveRecord::Base
 
   #-------------------
   # Associations
+  has_many :eloqua_emails
```
ewr: Is this really a one-to-many relationship? Do BreakingNewsAlerts get multiple emails created?

app/models/eloqua_email.rb
```diff
@@ -61,8 +56,141 @@ def obj_name
...
+  #
+  # Returns Hash of Strings
+  def eloqua_config config=nil
+    attributes = config || Rails.configuration.x.api.eloqua.attributes
+    (self.obj_name ? attributes[self.obj_name.underscore] : nil) || raise_missing_eloqua_config(obj_name)
+  end
+
+  def increment_attempts_made
+    # If the attempt number is nil, we do nothing.
+    if respond_to?(:attempts_made) && number = self.attempts_made
```
ewr: In our model, we should know whether this attribute exists and shouldn't have to test for it.

app/models/eloqua_email.rb
```diff
@@ -61,8 +56,141 @@ def obj_name
...
+  # Publish an e-mail for this object to the Eloqua API.
+  #
+  # Returns nothing.
+  def publish_email(options={})
+    return if !should_send_email?
+
+    if (email=create_email) && (campaign=create_campaign(email))
+      update_email_status(campaign)
+    end
+    try do
```
ewr: Why is this in a try?

ewr:
I made some specific comments, but generally I like where this is going.

I think there are some function names that could use a cleanup now that EloquaEmail is just implementing its functionality internally (`should_send_email`, etc). Similarly, status should probably be a function of whether the campaign and email were successfully created.

It would be nice to extend the EloquaEmail model to store the email and campaign IDs once they are generated, so that we could potentially give you a link from the CMS into Eloqua.

app/models/eloqua_email.rb
Ravenstine: Another relic from when I was writing this in the module.  It's not needed.

Ravenstine:
Won't fix.  Closing until we decide that we need this.

PR #391, opened 2015-12-02
Improve segment performance

ewr:
(Creating PR to make a few comments)

### DIFF
--- a/app/views/programs/kpcc/_featured_programs.html.erb
+++ b/app/views/programs/kpcc/_featured_programs.html.erb
@@ -5,28 +5,28 @@
       <% @featured_programs.each do |program| %>
         <% cache ["show", "featured-programs", "v1", program] do %>
           <% latest_episode = program.episodes.published.first %>
-            <article>
-              <figure class="photo ratio">
-                <div class="fill"></div>
-                <b class="img-contain"><a href="<%= program.public_path %>"><%= image_tag "programs/#{program.slug}.jpg" %></a></b>
-              </figure>
-              <a href="<%= program.public_path %>">
-                <h1><%= program.title %></h1>
-                <h2>With <%= program.host %></h2>
-                <h3><%= program.airtime %></h3>
-                <p><%= program.teaser.html_safe %></p>
-              </a>
-              <% if latest_episode.present? %>
-                <% cache ["show", "featured-programs", "latest-episode", "v1", latest_episode] do %>
-                  <a href="<%= latest_episode.public_path %>">
-                    <dl>
-                      <dt><mark>Latest</mark> <time datetime="<%= format_date(latest_episode.air_date, with: '%F') %>"><%= format_date(latest_episode.air_date, with: '%b. %-d') %></time></dt>
-                      <dd><%= latest_episode.headline %></dd>
-                    </dl>
-                  </a>
-                <% end %>
+          <article>
+            <figure class="photo ratio">
+              <div class="fill"></div>
+              <b class="img-contain"><a href="<%= program.public_path %>"><%= image_tag "programs/#{program.slug}.jpg" %></a></b>
+            </figure>
+            <a href="<%= program.public_path %>">
+              <h1><%= program.title %></h1>
+              <h2>With <%= program.host %></h2>
+              <h3><%= program.airtime %></h3>
+              <p><%= program.teaser.html_safe %></p>
+            </a>
+            <% if latest_episode.present? %>
+              <% cache ["show", "featured-programs", "latest-episode", "v1", latest_episode] do %>
+                <a href="<%= latest_episode.public_path %>">
+                  <dl>
+                    <dt><mark>Latest</mark> <time datetime="<%= format_date(latest_episode.air_date, with: '%F') %>"><%= format_date(latest_episode.air_date, with: '%b. %-d') %></time></dt>
+                    <dd><%= latest_episode.headline %></dd>
+                  </dl>
+                </a>
               <% end %>
-            </article>
+            <% end %>
+          </article>
         <% end %>
       <% end %>
     </div>
--- a/app/views/programs/kpcc/_featured_programs.html.erb
+++ b/app/views/programs/kpcc/_featured_programs.html.erb
@@ -3,9 +3,9 @@
   <div class="inner bound">
     <div class="troupe clearfix">
       <% @featured_programs.each do |program| %>
-        <% cache program do %>
+        <% cache ["show", "featured-programs", "v1", program] do %>
           <% latest_episode = program.episodes.published.first.try(:to_article) %>
-          <% cache latest_episode do %>
+          <% cache ["show", "featured-programs", "latest-episode", "v1", latest_episode] do %>
             <article>
               <figure class="photo ratio">
                 <div class="fill"></div>
--- a/app/views/programs/kpcc/_featured_programs.html.erb
+++ b/app/views/programs/kpcc/_featured_programs.html.erb
@@ -5,7 +5,6 @@
       <% @featured_programs.each do |program| %>
         <% cache ["show", "featured-programs", "v1", program] do %>
           <% latest_episode = program.episodes.published.first %>
-          <% cache ["show", "featured-programs", "latest-episode", "v1", latest_episode] do %>
             <article>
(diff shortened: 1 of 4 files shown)


app/views/programs/kpcc/_featured_programs.html.erb
```diff
@@ -3,29 +3,32 @@
...
-          </a>
-          <% if latest_episode.present? %>
-            <a href="<%= latest_episode.public_path %>">
-              <dl>
-                <dt><mark>Latest</mark> <time datetime="<%= format_date(latest_episode.public_datetime, with: '%F') %>"><%= format_date(latest_episode.public_datetime, with: '%b. %-d') %></time></dt>
-                <dd><%= latest_episode.short_title %></dd>
-              </dl>
-            </a>
+        <% cache ["show", "featured-programs", "v1", program] do %>
+          <% latest_episode = program.episodes.published.first.try(:to_article) %>
```
ewr: (this isn't something you added, but) `to_article` is evil and should only be used where necessary, as close as possible to the spot where it is necessary. In this case we're using the episode's timestamp, path and title, none of which need an Article object.

app/views/programs/kpcc/_featured_programs.html.erb
```diff
@@ -3,29 +3,32 @@
...
+          <% cache ["show", "featured-programs", "latest-episode", "v1", latest_episode] do %>
```
ewr: This cache should only enclose the template code that deals with the episode (lines 20 - 27 or so).

app/views/programs/kpcc/_recent_segments.html.erb
```diff
@@ -1,48 +1,49 @@
...
-                <time datetime="<%= format_date(segment.published_at, with: "%F") %>"><span><%= format_date(segment.published_at, with: "%a") %></span> <%= format_date(segment.published_at, with: "%B %e") %></time>
-                <% if segment.audio.any? %>
-                  <mark class="has-audio"><%= format_clip_duration segment.audio.first.duration %></mark>
-                <% end %>
-                <b class="img-contain"><img src="<%= segment.asset.full.url %>" alt="" /></b>
-              </a>
-            </div>
-            <figcaption><a href="<%= segment.public_path %>"><%= segment.short_headline %></a></figcaption>
-        </figure>
+  <% cache ["featured-program-segments","program",program] do %>
```
ewr: "program" as a string is unnecessary in this key.

app/views/programs/kpcc/_recent_segments.html.erb
```diff
@@ -1,48 +1,49 @@
...
+    <div class="series clearfix">
+      <% segments.first(14).each do |segment| %>
+        <% cache ["featured-program-segments",segment] do %>
```
ewr: I doubt there's a major win from caching here given the new cache outside it. Worth profiling, though. Basically you're interested in render time when the program is updated but the segments are cached vs render time if program is updated and there's no inner cache.

app/views/programs/kpcc/_recent_segments.html.erb
```diff
@@ -1,48 +1,49 @@
...
+          <figure>
+              <div class="photo ratio">
+                <a href="<%= segment.public_path %>">
+                  <div class="fill"></div>
+                  <time datetime="<%= format_date(segment.published_at, with: "%F") %>"><span><%= format_date(segment.published_at, with: "%a") %></span> <%= format_date(segment.published_at, with: "%B %e") %></time>
+                  <% if segment.audio.any? %>
```
ewr: Because we check for `.audio.any?` on each segment here, it would be worth profiling what it looks like to include the audio objects in the fetch in `programs_controller#featured_program`. Basically `.includes(:audio)` on the definition for `@segments` before we hit the kicker in the template.

app/views/programs/kpcc/_recent_segments.html.erb
Ravenstine: This did reduce the response time by around 20ms. (from 85ms to about 65ms)  Switched back to how it was before just to be sure.  Not sure if that's what we'd consider a major win.

I think that might be due to the fact that it isn't looping through up to 14 different segment caches.  Since a change to a segment should touch the program, it appears to me that it might work just fine since the segments that didn't change should still have their own caches the next time the program cache fragment decides to update.

ewr: Right. I'm just saying with the outer cache, the nested inner cache per-segment probably doesn't help much. Obviously when the program cache is in place, you'll just get that, so you're interested in the render times for the load after a program is touched.

Ravenstine: Ah, I get ya.

app/views/programs/kpcc/_recent_segments.html.erb
Ravenstine: Still working on this one, though I've addressed your other comments.

Ravenstine: Including the audio on the definition for `@segments` doesn't seem to have a perceptible effect.

ewr: Two things:


  1.  You’re talking about during a case where you don’t have a cache, correct? Obviously it won’t make a different in the cached case.
  2.  Does it make a difference to the database calls being generated? In the current case, you have one query to get the segments, and 14 queries to load audio. In the includes case, you should have two SQL calls.

sdillingham:
@Ravenstine I think this PR is close to being ready to deploy, but there was one unfinished thread in the discussion between you and @ewr regarding the `.audio.any?` check in the `_recent_segments` partial. Once you close the loop on that conversation and may any relevant changes, I think you've got some good performance wins here that would be :shipit: -able.

app/views/programs/kpcc/_recent_segments.html.erb
Ravenstine: I probably had caching turned on.  With it explicitly off, including the audio in the `@segments` definition shaves off about 20ms.

The difference seems to be that it's doing the following, when we're using `#includes`:
```
SELECT `media_audio`.* FROM `media_audio` WHERE `media_audio`.`content_type` = 'ShowSegment' AND `media_audio`.`content_id` IN (45400, 45408, 45402, 45399, 45398, 45396, 45394, 45393, 45392, 45389, 45388, 45377, 45376, 45383)  ORDER BY position
```

As opposed to doing this for every single segment:
```
  Audio Exists (0.2ms)  SELECT  1 AS one FROM `media_audio` WHERE `media_audio`.`content_id` = 45398 AND `media_audio`.`content_type` = 'ShowSegment' LIMIT 1
  Audio Load (0.2ms)  SELECT  `media_audio`.* FROM `media_audio` WHERE `media_audio`.`content_id` = 45398 AND `media_audio`.`content_type` = 'ShowSegment'  ORDER BY position LIMIT 1
```

Ravenstine:
@sdillingham Sorry, I didn't get back to that conversation sooner.  I made a comment in that discussion and I've also pushed a relevant change.  I'll leave this here for now until @ewr gets back tomorrow and we can :shipit: .

app/views/programs/kpcc/_recent_segments.html.erb
Ravenstine: @ewr Are we ready to ship, or is there anything else we need to address?

app/views/programs/kpcc/_featured_programs.html.erb
```diff
@@ -3,29 +3,32 @@
...
-              <dl>
-                <dt><mark>Latest</mark> <time datetime="<%= format_date(latest_episode.public_datetime, with: '%F') %>"><%= format_date(latest_episode.public_datetime, with: '%b. %-d') %></time></dt>
-                <dd><%= latest_episode.short_title %></dd>
-              </dl>
-            </a>
-          <% end %>
-        </article>
+        <% cache ["show", "featured-programs", "v1", program] do %>
+          <% latest_episode = program.episodes.published.first %>
+            <article>
```
ewr: `latest_episode` isn't a block, so the bits below it shouldn't be indented from it.

ewr:
Other than the one formatting comment, this looks good to :shipit:

Issue #381, opened 2015-11-13
Refine Tag Implementation


sdillingham:
- [x] Remove "is_featured" attribute from Tag model/Issues controller + view, plus outpost filters and forms
- [x] Add `tag_image` attribute (maybe just use asset ID field like we do in Bios?)
- [x] Add `tag_type` attribute with three values ("beat", "series", "keyword").
- [x] Add Outpost UI to support groupings of tags by type in the tag dropdown.
- [x] Delete "Issues" layout and have issues use "Ronin" layout
- [x] Add "parent entity" attribute to Tag model. Allow KPCC Programs to claim tags that belong to them.

Ravenstine:
Will `tag_type` be a required field or have a default value?

sdillingham:
Good question - it should be required for sure, and I think "Keyword" is the right default value. "Series" and "Beat" are tag types that have greater implications for the design and front-end experience, so I think Outpost users should have to be more intentional about selecting those.

Ravenstine:
`Add Outpost UI to support groupings of tags by type in the tag dropdown.`

I am trying to envision how this would be implemented.

The simplest solution I can think of is to add the grouping name between brackets `i.e [series] Higher Education` to the dropdown search so that a person can search by grouping and/or the name of the tag.  It would also let them see the grouping type along with each tag.

But I'm guessing we have something else in mind?  Maybe multiple columns?

sdillingham:
I think we keep it simple. Select2, the jQuery plugin Outpost depends on for its dropdown functionality, has the ability to transform <optgroup> collections in a standard <select> element and create headings out of them:

https://select2.github.io/examples.html#single

I think that's sufficient for an MVP.

Ravenstine:
:+1:

Ravenstine:
@sdillingham As far as programs having ownership of tags, do we need the ability to have tags belong to multiple shows or is it fine to just give tags a foreign key to a program(i.e. program has many tags, tag has one program)?

sdillingham:
I think @ewr had the idea to make this a polymorphic association so that we have an interface that any other model can use to be a parent entity for tags (correct me if I'm wrong). We'd start with KPCCPrograms, but we want the flexibility for other models to act as parent entities in the future. 

I'm in favor of tags only being able to have a single parent entity. I think things will get confusing if a series could belong to multiple parent entities, and I think it's unlikely that we'll need that capability.

Ravenstine:
[...]

PR #392, opened 2015-12-03
Refine tag implementation

Ravenstine:
#381

One part of this I think might need to be changed is the way that the `tag types` are implemented.  Initially, I went with just a string value but convinced myself that this needed to be a relation because of the way that simple_form creates the select field(which expects a collection of type objects that contain their respective tags).  But now that I've gotten to this point, I believe this is likely unnecessary since it doesn't seem like we are changing these types often(or at all?).  This would be a simple change, especially since that's how it worked in earlier commits.

While I added the ability to select a parent program from the Tag edit view, I also left the tag association selection in the Program edit view.  I'm not sure if we need to get rid of the latter, but we certainly could if that seems too hazardous.

I removed the `pry-byebug` gem because not only is it not necessary but it often slows down the development environment significantly.


Ravenstine:
@sdillingham @ewr Do you guys have an opinion on whether or not the tag groups should just be string fields or should be associative?  

The only reason I ask is because I ended up changing it to be associative to fulfill the way that simple_form creates grouped selects, even though it can be worked around.

`f.association :tags, as: :grouped_select, collection: TagType.all, group_method: :tags, input_html: { :multiple => true }`

sdillingham:
I'm not sure I see the benefit to tag type being its own model with an association. Why not use a constant like this:

https://github.com/SCPR/SCPRv4/blob/master/app/models/news_story.rb#L43-L55

sdillingham:
In general I don't want to make decisions about how we model data that are influenced by how things like simple_form behave.

Ravenstine:
Cool, I think I agree. :+1:

Ravenstine:
@sdillingham So I've made some changes based on my last question, and I've put the work up on staging.  I have also gotten rid of the tag selection in the KpccProgram edit view since it does seem better to just select the parent in the Tag view.

Ravenstine:
@sdillingham @ewr There's some peculiar behavior that's left me stumped.

On my machine, both the `tag_type` and `parent` fields save properly upon edit.

Yesterday, on Sean's machine, neither field would save.  No failure would occur; the values just don't persist.

On staging, for some weird reason, the migrations seemed to have run but the database still had an older field name for `tag_type` (which was `tag_type_id`).  A migration takes care of that name change but even when I rolled back and re-migrated, I'd get the same issue.  But I'm digressing a little, since that problem has been resolved.

Now that the code is now running on staging, the `tag_type` field never seems to persist, but the `parent` field seems to persist about 60% of the time.

I can't figure out why this would be, since the changes to make these features happen were adding the fields to the table, defining the columns in `Outpost::TagsController`, and using the form helpers in the view to create the right select tags.  The option tags in the select appear to have the correct values.

I'm not sure what's going on here.

Issue #383, opened 2015-11-13
Transition "Issues" to "Topics"


sdillingham:
- [x] simplify tag page design to remove all the "Issues"-specific UI cruft (Needs comps from design)
- [x] update routes to support both `/issues/<tag-slug>` (deprecated) and `/topics/<tag-slug>`
- [x] have `/issues/` redirect to homepage(?)
- [x] Change "More from this Issue" sidebar heading to "More from this Topic"
- [x] Change "Issues We're Tracking" heading on Vertical pages to "Topics We're Tracking" and remove "See All Issues" link
- [x] Rename "Issues"-related controllers and views to "Topics"

Ravenstine:
@sdillingham Since we are now distinguishing our tags between `Keyword`, `Beat`, and `Series`, are we still going to have our Issues(now Topics) page include all tags?  Or do we want to restrict it to a type(s)? (e.g. keywords and beats)

sdillingham:
TL;DR - let's have our topic pages include all tags.

As part of rolling out this change, We'll redesign these topic pages. To start, we'll remove the list of tags down the left side of the page. A second phase will probably involve creating different templates for each tag type, but we haven't defined specific requirements for that yet.

Ravenstine:
Great, I think I can work on the other criteria of the AC and integrate the new designs later.

Ravenstine:
@sdillingham On staging, I have made the changes that I have check-marked from the AC.  Here's a pull request to see the changes in code:
https://github.com/SCPR/SCPRv4/pull/393

Barring any fixes I might need to make to the tests(or possibly stylesheets), it seems like the only main thing this is waiting on is comps from design for the simplified tag page.

Ravenstine:
Also, the problem of the tag attributes not persisting on save seems to be gone.  I'm guessing it was an issue with the schema not loading properly.  But it seems to no longer be an issue.

sdillingham:
Thanks for the heads up @Ravenstine, I'll check it out on staging.

Issue #382, opened 2015-11-13
Revise Taxonomy


sdillingham:
- [x] Populate "beat" tags for each of the regional desk beats.
- [x] Migrate all stories categorized as Immigration to a tag and delete the Immigration Category from the DB and instances of the site nav.

Ravenstine:
@sdillingham Do we have a list of these beat tag names?

sdillingham:
Yep, here they are:

* Public Safety
* SoCal Economy
* SoCal Politics
* The Social Safety Net
* Vets
* Workplace
* Changing Neighborhoods & Affordability
* Communting
* Immigration 3.0
* Infrastructure
* Orange County

Ravenstine:
I'm assuming that I can make up the slugs for these.  I will probably need descriptions for each of them at some point too.

sdillingham:
Yeah, descriptions can come later. Here's slugs you can use:

public-safety
socal-economy
socal-politics
social-safety-net
vets
workplace
changing-neighborhoods
communting
immigration
infrastructure
orange-county

PR #393, opened 2015-12-07
Transition issues to topics

Ravenstine:
Takes code yet to be deployed from #381 and addresses #383.

Removes the idea of "Issues" and moves anything related over to "Topics".


sdillingham:
@Ravenstine I think #383, #382 and #381 make sense to bundle together as a single release (I rolled them up into a milestone here: https://github.com/SCPR/SCPRv4/milestones/Content%20Taxonomy%20Overhaul). So while it's fine for reviewing and commenting on work to issue these as separate pull requests, we'll want to merge this work together into a single pull request prior to merging into master and deploying.

sdillingham:
@Ravenstine One minor change: I do think we should remove this link:

https://github.com/SCPR/SCPRv4/blob/transition_issues_to_topics/app/views/verticals/shared/_lunchbox.html.erb#L9

We're going to have too many tags for us to build a top-level topics page that lists all our topics in one place, and I doubt that page would be of much value to the audience. So I'd rather start getting rid of these links now in anticipation of that change.

PR #394, opened 2015-12-07
Content Taxonomy Overhaul #1

Ravenstine:
This addresses issues #383, #382 and #381, which are focused on migrating away from Issues to different forms of Tags that represent keywords, beats, and series.  I originally had pull requests for review based on the separate tickets, but now I'm rolling that work into one pull request.


sdillingham:
@Ravenstine this is all looking good so far. This deploy may require som communication with the station to explain some of the new features, so @vsinghkpcc and I are going to review this week and come up with a communication plan. Once we have a plan and a target launch date we can merge and deploy this.

db/migrate/20151207192853_move_immigration_stories_to_immigration_tag.rb
```diff
@@ -0,0 +1,20 @@
+class MoveImmigrationStoriesToImmigrationTag < ActiveRecord::Migration
+  def up
+    category = Category.where(slug: 'immigration').first!
+    story_ids = NewsStory.where(category_id: category.id).pluck(:id)
```
sdillingham: We need to also include BlogEntry, ShowSegment and Events content in this migration, or else we'll wind up with a bunch of objects with an orphaned category_id. Abstracts also have a Category association, but can't be tagged, so with those we should just clear out the `category_id` value. There are some other spots in the DB that will need cleanup, but I'll add those as a comment on the PR.

sdillingham:
Since the Immigration Category is going away, that will result in the Immigration Featured Comment Bucket being orphaned. That bucket is also associated with a bunch of Featured Comments.

Given that Featured Comments are only used on the homepage, and the homepage design is changing soon, these models may soon be obsolete. I spoke with Eric Z. and he is OK with destroying both the Immigration Bucket and the Featured Comments associated with the Immigration Bucket. The content contained in those objects is old, aggregates quotes from content that lives elsewhere, and isn't likely to ever be used again.

Ravenstine:
I added a migration that removes the featured comment bucket for the immigration category, as well as its associated comments.  I also changed the previous migration for populating the beat tags to go through each of the tables that has a category_id associated with contentbase_category.

app/models/tag.rb
```diff
@@ -8,6 +8,10 @@ class Tag < ActiveRecord::Base
 
   has_many :taggings, dependent: :destroy
 
+  belongs_to :parent, polymorphic: true
+
+  TYPES = ["Keyword", "Series", "Beat"]
```
sdillingham: In the Outpost dropdown, it would make more sense for this to be in the opposite order, since KPCC reporters will most often be filing content that is associated with their beat: 

["Beat", "Series", "Keyword"]

sdillingham:
@Ravenstine, two more small changes:

These references to the "/immigration" category page need to be removed from the masthead and footer nav:

- https://github.com/SCPR/SCPRv4/blob/populate_beat_tags/app/views/shared/masthead/_news.html.erb#L24
- https://github.com/SCPR/SCPRv4/blob/populate_beat_tags/app/views/shared/footer/_layout_footer.html.erb#L55

Ravenstine:
:point_up_2: I removed those links as well as one in the old layout footer.  I thought why not.

Issue #370, opened 2015-11-02
Need a Periodic Job to Review Schedule Occurrences for Overlaps and Gaps


ewr:
Over the weekend, a gap at the time of the time change caused iPhone app clients to attempt to DDOS the website using schedule endpoints. While the iPhone app schedule logic is a bug, this would have been avoided had we recognized the gap ahead of time.

AC

- [ ] We need a periodic (daily?) job that reviews the upcoming schedule to identify overlaps and gaps and emails a warning to the correct parties.

NOTE: Vijay will help identify correct parties, and those people will be in an e-mail group

Ravenstine:
@vsinghkpcc Is there a specific email group in Eloqua that I should have this job send notifications to? (or if not, who should we include in an email group?)

vsinghkpcc:
Submitted a request this morning for a new e-mail group to send this notification, just waiting to hear back...

vsinghkpcc:
@Ravenstine We have the e-mail group: digitalbroadcastopsscpr@scpr.org

PR #395, opened 2015-12-08
Check for problems in program schedule

Ravenstine:
#370

This adds a job that looks for gaps and overlapping in ScheduleOccurrences and sends an email reporting them to an email address.

Some things like the email template as well as the destination address may need changing.  Right now, the destination address is fake.


app/mailers/report_schedule_problems_mailer.rb
```diff
@@ -0,0 +1,9 @@
+class ReportScheduleProblemsMailer < ActionMailer::Base
+  def send_notification gaps, overlaps
+    @gaps     = gaps
+    @overlaps = overlaps
+    mail(to: "scprscheduleproblems@safetymail.info",
```
sdillingham: We'll need to identify the correct group email address to send to so this reaches the right individuals. @vsinghkpcc can you help with this?

app/views/report_schedule_problems_mailer/send_notification.html.erb
```diff
@@ -0,0 +1,22 @@
+There seems to be a few problems with the current SCPR program schedule:
```
sdillingham: Minor grammar issue: change "seems" to "seem"

app/views/report_schedule_problems_mailer/send_notification.text.erb
```diff
@@ -0,0 +1,17 @@
+There seems to be a few problems with the current SCPR program schedule:
```
sdillingham: same grammar change here

app/views/report_schedule_problems_mailer/send_notification.html.erb
Ravenstine: Thanks. :wink:

sdillingham:
Now that we have multiple mailers, I'd like to lobby for reorganizing mailer views into a common directory at `app/views/mailers` (i.e. `app/views/mailers/content_mailer`, `app/views/mailers/report_schedule_problems_mailer`). It's unwieldy to have `content_mailer` and `report_schedule_problems_mailer` directories as subdirectories inside `views`, and it will only get worse if we add more in the future.

You could create an `ApplicationMailer` class with your defaults and inherit from that in your mailers, and should be able to set up your custom view path there.

Ravenstine:
:thumbsup:

app/mailers/report_schedule_problems_mailer.rb
```diff
@@ -0,0 +1,8 @@
+class ReportScheduleProblemsMailer < ApplicationMailer
+  def send_notification gaps, overlaps
+    @gaps     = gaps
+    @overlaps = overlaps
+    mail(to: "scprscheduleproblems@safetymail.info",
```
ewr: This should be a configuration setting instead of being hardcoded in the mailer.

app/models/schedule_occurrence.rb
```diff
@@ -100,6 +100,29 @@ def block(date, length, collapse=false)
 
       occurrences
     end
+
+    def gaps
+      gaps = []
+      future.order("starts_at ASC").each_cons(2) do |pair|
```
ewr: Isn't that order already the one being applied by the scope?

Ravenstine:
@ewr I had initially written it that way(to return a hash with both).  I decided to split it up because, since we're only looking ahead through about 2 weeks of occurrences at most, the performance was fast either way.  It was an arbitrary decision.  If you think it'd be better to make it one, I have no arguments against that.

ewr:
To me they're a single concept—an aberration in the schedule. I would create them both in the same pass, and would present them together.

app/mailers/report_schedule_problems_mailer.rb
Ravenstine: Correct.  Waiting on the actual email address.

Issue #396, opened 2015-12-09
Need to support DAAST Format for Ad Response


ewr:
Currently ads use an overloaded VAST for audio. We're working on switching our ad infrastructure over to DAAST, which is similar but designed for audio. 

Current ad parsing implementation is here:

https://github.com/SCPR/SCPRv4/blob/8f02a64b3905e8b94fb8db4c4c81f03437653b15/app/assets/javascripts/listen_live.js.coffee#L116-L164

I'm not sure if there's a way to reuse StreamMachine's ad parsing code (also written in Coffeescript), but it's here: 

https://github.com/StreamMachine/StreamMachine/blob/2d819379fa2cdc9505d9c3cc20663b0731d68be5/src/streammachine/slave/preroller.coffee#L234-L328

* [x] Preroll should parse and handle either VAST or DAAST responses, without knowing which one it will receive.
* [x] Preroll should support multiple impression URLs, as allowed by the spec.
* [x] Preroll should correctly hit the companion ad's creativeView tracking URL when creative is displayed.

ewr:
Test preroll endpoint is:

http://adserver.adtechus.com/?adrawdata/3.0/5511.1/3683959/0/1013/header=yes;cookie=no;adct=text/xml

There should always be a dev campaign running.

ewr:
Actually, if the display is using the html version of the companion ad, creativeView should already be covered via a `display:none` image. Take a look at the XML response and you'll see what I'm talking about.

Ravenstine:
@ewr It looks like the test endpoint you posted disappeared somehow.  Can you re-add it?

ewr:
Works for me...

Ravenstine:
I've been reading the DAAST spec.  For clarity, are we essentially wanting to do the same thing with both the impression URLs and the creativeView tracking URL?  It seems like all we need to do is just touch them.

Also, between lines 58 and 64 in listen_live.js.coffee, it appears like we have some commented out code that is intended to do the touching for a single impression URL:
```coffee
                if url = @_impressionURL
                    @_impressionURL = null
                    # create an img and append it to our DOM
                    img = $("<img src='#{url}'>").css("display:none")
                    @_live_ad.append(img)
                    #$.ajax type:"GET", url:"#{url};cors=yes", success:(resp) =>
                        # nothing right now
```

If we are expecting multiple URLs, how do we want to handle the portion where an image is being added to the DOM?  Do we want to add one for each impression URL?

[...]

PR #403, opened 2015-12-28
Add DAAST support to live stream

Ravenstine:
Issue #396 

Intended to add support in the live stream code for both DAAST and VAST responses for preroll.

Note that lines 117 - 119 are incorrect and that will be fixed soon.

### DIFF
--- a/app/assets/javascripts/listen_live.js.coffee
+++ b/app/assets/javascripts/listen_live.js.coffee
@@ -1,10 +1,6 @@
 #= require xml2json
 
 class scpr.ListenLive
-    constructor: (options) ->
-        # This was old code for a StreamMachine PoC. It is no longer used.
-
-    #----------
 
     class @CurrentGen
         DefaultOptions:
@@ -19,20 +15,23 @@ class scpr.ListenLive
             skip_preroll:       false
 
         constructor: (opts) ->
-            @options = _.defaults opts, @DefaultOptions
+            @options        = _.defaults opts, @DefaultOptions
 
-            @x2js = new X2JS()
+            @x2js           = new X2JS()
 
-            @player = $(@options.player)
+            @player         = $(@options.player)
+
             @_pause_timeout = null
 
-            @_live_ad = $(@options.ad_element)
+            @_live_ad       = $(@options.ad_element)
 
-            @_shouldTryAd = true
+            @_shouldTryAd   = true
 
-            @_on_now = null
+            @_inPreroll     = false
 
-            @_playerReady = false
+            @_on_now        = null
+
+            @_playerReady   = false
 
             # -- set up our player -- #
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/listen_live.js.coffee
```diff
@@ -131,35 +132,63 @@ class scpr.ListenLive
...
-                                @player.jPlayer("setMedia",mp3:media).jPlayer("play")
-
-                                @_impressionURL = ad.Impression.toString()
-
-                            else
-                                _playStream()
-
-                            # is there a visual?
-                            if companions = ad.Creatives?.Creative?[1]?.CompanionAds?.Companion
+                        class AdResponse
```
ewr: This shouldn't be sitting in the middle of a function. It's fine to have it sit inside the ListenLive class.

app/assets/javascripts/listen_live.js.coffee
```diff
@@ -131,35 +132,63 @@ class scpr.ListenLive
...
+                                # give DAAST higher precedence than VAST
+                                @obj = obj.DAAST or obj.VAST or undefined     
+                            preroll: ->
+                                @ad().Creatives?.Creative?[0]?.Linear
+                            ad: ->
+                                @obj?.Ad?.InLine
+                            renderVisual: (el)->
                                 # find the first html or iframe
-                                _(companions).find (c) =>
+                                _(@companions()).find (c) =>
```
ewr: I know this logic isn't code you wrote, but we probably want to have a preference order for creatives. I think this loop would just return the first match. HTMLResource -> IFrameResource -> StaticResource would be a logical order, even though currently we will always return an HTMLResource if there's an ad.

app/assets/javascripts/listen_live.js.coffee
```diff
@@ -131,35 +132,63 @@ class scpr.ListenLive
...
+                                _.map parsedImpressions, (i) ->
+                                    return i.toString()
+                            companions: ->
+                                @ad().Creatives?.Creative?[1]?.CompanionAds or @ad().Creatives?.Creative?[1]?.CompanionAds?.Companion or []
+                            playPreroll: (player)->
+                                if preroll = @preroll()
+                                    media = preroll.MediaFiles.MediaFile.toString()
+                                    player.jPlayer("setMedia",mp3:media).jPlayer("play")     
+                            _submitViewEvent: (companion) ->
+                                if url = companion.TrackingEvents.Tracking.toString()
```
ewr: There can be other tracking events besides `creativeView`. For instance, in the video world you would typically send events for quartiles played, etc.

app/assets/javascripts/listen_live.js.coffee
```diff
@@ -131,35 +132,63 @@ class scpr.ListenLive
...
+
+                        response = new AdResponse(obj)
 
+                        # is there an ad there for us?
+                        if response.ad()
+                            # is there a preroll?
+                            if response.preroll() && !_timedOut
+                                # yes... play it
+                                response.playPreroll(@player)
+                                @_impressionURLs = response.impressions()
```
ewr: Instead of stashing the impression URLs here, it would make sense to just stash the AdResponse object and then use it directly when it comes time to hit the impression URL(s). That makes life a little easier for debugging.

PR #405, opened 2015-12-29
Integrate Nielsen SDK

Ravenstine:
Includes the Nielsen SDK and reports plays, stops, and playhead position(periodically) while our content plays through the live stream player.

This also includes work I have done to support DAAST, which is now in production.


app/assets/javascripts/listen_live.js.coffee
```diff
@@ -55,21 +51,19 @@ class scpr.ListenLive
...
                     @_pause_timeout = null
 
-                if url = @_impressionURL
-                    @_impressionURL = null
-                    # create an img and append it to our DOM
-                    img = $("<img src='#{url}'>").css("display:none")
-                    @_live_ad.append(img)
-                    #$.ajax type:"GET", url:"#{url};cors=yes", success:(resp) =>
-                        # nothing right now
+                @adResponse?.touchImpressions(@_live_ad)
```
ewr: Totally minor, but I'm not sure there's any need for these images to be inserted by the ad in the DOM. I think the touch code could just append into the body element and not need to pass this around.

app/assets/javascripts/listen_live.js.coffee
```diff
@@ -226,4 +204,102 @@ class scpr.ListenLive
...
+                _.each impressions, (url) =>      
+                    # create an img and append it to our DOM
+                    img = $("<img src='#{url}'>").css("display:none")
+                    el.append(img)
+
+    #----------
+
+    class Nielsen
+        constructor: ->
+            if NOLCMB?
```
ewr: Would it make sense for the nielsen script require to actually be injected here? That allows it to only be injected when needed, and for you to know that it succeeded / failed directly rather than just via hoping it's defined.

ewr:
I don't love defining / undefining the nielsen object based on whether it's a preroll playing. I'd rather see it defined when the player is initialized, and then just better logic around when to call play and stop defined.

That could be something like a `_isPreroll` boolean guard that gets set by the code to load the preroll and unset on the subsequent trip through `_playStream`. Then it should be pretty simple logic to just call play / stop if `!@_isPreroll`.

Ravenstine:
@ewr @sdillingham @vsinghkpcc 

I have addressed the comment left by @ewr above.  If this is ready, the next step is getting certification from Nielsen.  The documents I have mention this aren't too specific, so I'm guessing this involves some direct contact with them.

ewr:
Implementation of the Nielsen SDK requires us to add Nielsen's opt-out verbiage to our Privacy Policy. We currently include a Nielsen//NetRatings section, but that's a different product with a different privacy page. Not sure if it counts.

Sample verbiage from the Nielsen presentation:

> Our properties may feature Nielsen’s proprietary measurement software which will allow you to contribute to market research, like Nielsen’s TV Ratings. To learn more about the information Nielsen’s software may collect and your choices with regard to it, please click here [http://www.nielsen.com/digitalprivacy].

app/assets/javascripts/listen_live.js.coffee
```diff
@@ -54,14 +56,17 @@ class scpr.ListenLive
...
+
+                @nielsen?.play() if !@_inPreroll
 
             @player.on $.jPlayer.event.pause, (evt) =>
                 # set a timer to convert this pause to a stop in one minute
                 @_pause_timeout = setTimeout =>
                     @player.jPlayer("clearMedia")
                     @_shouldTryAd = true
-                , @options.pause_timeout * 1000
+                , @options.pause_timeout * 10
```
ewr: I assume you didn't mean to check this in.

ewr:
I would still be in favor of requiring the Nielsen script directly in ListenLive.Nielsen, not separately. It keeps the code in one place. A slightly-refactored implementation might look like:

```coffee
    class Nielsen
        constructor: ->
            @_queued = []
            $.getScript "http://secure-cert.imrworldwide.com/novms/js/2/ggcmb400.js"
                .done (script,status) =>
                    @nolcmb = new NOLCMB?.ggInitialize
                        sfcode: "uat-cert"
                        apid  : "T4FA39C01-1BC0-41C3-A309-06ED295D84D2"
                        apn   : "test"

                    @nolcmb.ggPM e... for e in @_queued
                    true

                .fail (xhr,settings,exception) =>
                    console.log "Failed to load Nielsen SDK: #{exception}"

        _send: (event,data) ->
            if @nolcmb
                @nolcmb.ggPM event, data
            else
                @_queued.push [event,data]

        play: ->
            @_send "loadMetadata",
                stationType: 2
                dataSrc    : "cms"
                type       : "radio"
                assetid    : "KPCC-FM"
                provider   : "Southern California Public Radio - Radio"

            @_send "play", Math.floor(Date.now() / 1000)

            # send setPlayheadPosition every 2 seconds, as specified by Nielsen
            @_setPlayheadPosition = setInterval(=>
                @_send "setPlayheadPosition", Math.floor(Date.now() / 1000)
            , 2000)

        stop: ->
            @_send "stop", Math.floor(Date.now() / 1000)
            clearInterval(@_setPlayheadPosition)
```

Ravenstine:
I thought those both allowed for the same opt-out, but they are different privacy policies so I think it doesn't count and we have to include that paragraph.  Would it be fine to append that paragraph to the `Nielsen//NetRatings` section?

Issue #409, opened 2016-01-14
Backend adjustments for better breaking news alerts in apps


vsinghkpcc:
AC
- [x] An approach for setting up breaking news alerts / schedule updates in outpost is defined and the newsroom is ok with the defined approach
- [x] The adjustments have been made in outpost
- [x] API updated to include additional data ("start time", "end time", "Should rewind", "bypass pre-roll")
- [x] Payload updated to include additional data
- [x] The backend has a way to notify clients when a breaking news alert ends

sdillingham:
Are we tracking the changes we'll need to make to the API and our Parse pushes in a separate ticket, or should those be part of the AC for this ticket as well?

Ravenstine:
Would I be correct in assuming that there will be no such concept as a "recurring rule" for breaking news alerts?  Of course it may be obviously contradictory, but I want to make sure my understanding is in line with what needs to be accomplished.

sdillingham:
No, there is no concept of recurrence for breaking news alerts. Also, in sprint planning we discussed our options, and decided not to create schedule occurrence objects as part of how we solve this problem. The idea is to add attributes to the Breaking News model only; it will be up to the client (e.g. the iPhone app) to take the metadata passed in a breaking news alert and do what needs to be done to treat it like a schedule occurrence in the UI.

Ravenstine:
:+1:

Ravenstine:
NOTE: Currently need to figure out whether or not we need to prevent the silent pushes from going to Android devices.

PR #422, opened 2016-02-02
Update Alert Push Notifications Through Silent Notifications

Ravenstine:
#409

Adds some UI to allow a CMS user to add meta data that would be useful for the mobile apps when they have to handle push notifications.  This includes `starts_at`, `ends_at`, `bypass_preroll`, `should_rewind`.  

The silent notifications target iOS devices that have a version of the app that is equal to or greater than the version that we specify, which would be a future build of the app that can properly handle the silent pushes.


Ravenstine:
Won't fix, for now.  Closing until we ever decide that we need this.

Issue #414, opened 2016-01-26
Need Strategy for Managing Size of Versions Table


ewr:
The `versions` table is growing unwieldy. For scpr.org, `versions` is now 591K rows and ~2550MB of data and indices.

Questions:
* Should version information be kept forever?
* Is there waste inside the way that version data is currently being stored that could be deleted without shortening the version history?

vsinghkpcc:
AC: Establish one or more strategies for solving this problem.

Ravenstine:
Currently looking into whether or not there's some waste being created in the way we are versioning that can help us reduce the problem.

Ravenstine:
@sdillingham @ewr

I have been taking a look at how we are versioning our content, and I haven't found that much room for improvement in terms of where we can reduce our waste.  We are already saving only changes(i.e. diffs) between versions, so I can only really see considering whether or not there are any fields that we don't want to save.  Whether or not smaller fields are dwarfing fields containing body text is a different question.  The size of our `news_story` table is only 322mb on my machine.

The `versions` table was over 2GB initially on my machine(unfortunately, I didn't record the exact number) but, after enabling compression on the table, I was able to reduce it down to 1063mb, which cut the size nearly in half.  

There is also the option of pruning the table and periodically storing the rest in the cloud(maybe Amazon S3).  Are we still considering something like this as a solution?

Who can I talk to in order to determine how far back we usually need to look at our versions data?  If we never look at anything from over 2 years ago, maybe version data that is over 2 years old can be periodically deleted and we don't need to bother saving any of it.

Ravenstine:
There may be some benefit to not saving versions unless a piece of content has been published.  Right now, we are creating versions even if a news story is in `draft`.

sdillingham:
@Ravenstine I spoke with Eric Zassenhaus and we tentatively agreed on a one-year retention policy, with the idea that we would dump out pruned records to a text file for archival storage. I still need to get buy in from others in the newsroom for 1 year as the right retention timeframe, but for now I think we can march forward with a process that periodically looks for versions more than 1 year old and writes them all to a flat file. Not sure what the best format for that data would be. JSON or CSV?

Ravenstine:
[...]

PR #423, opened 2016-02-03
Archive and delete old Secretary::Versions

Ravenstine:
#414 

This adds a task will create a zipped CSV file of old version records that are greater than a year old, upload the zip file to Amazon S3, and delete the old records from the table.


Ravenstine:
@sdillingham I didn't have any specific rationale for weekly.  I just picked it.  Monthly should be just fine.

Ravenstine:
I'll also make a note that I initially wrote this to do everything in memory, but switched to tempfiles because the results I was getting were taking up several megabytes(the first one in the GB range), and so tempfiles keep most of that out of memory as long as I also use `#find_in_batches`.  In case anyone was wondering.

sdillingham:
Yeah, let's try monthly. It does mean that we'll be dealing with larger file sizes every time we run the job, but we'll generate fewer files, and I think scoping files to an entire month means you'll have fewer instances of having to look across multiple files if you ever needed to reverse engineer the history of a particular record.

PR #424, opened 2016-02-03
Set active record default timezone to UTC

chris-fuller:
When running the app locally in development, and with a clone of the production database, all dates are 8 hours off.

This change fixes that issue. Also UTC is the local timezone on the server, so this ideally will not have any adverse effects in production.

However, this change should be tested **thoroughly** in staging before deploying.


Ravenstine:
@sdillingham @chris-fuller I put it on staging and it looks good to me.

sdillingham:
@Ravenstine can you point me to some spots in the code that would be affected by this change so I can verify myself that this looks good on staging?

Ravenstine:
@sdillingham The places that I looked at in particular are the `home_controller` where the schedule occurrences are grabbed for the broadcast bar, and the `edition` model where it makes the decision to send the *monday* short list.  I tested the logic in both by hand on the staging server and found nothing wrong.  I am still feeling wary of this, however.

sdillingham:
I'm wary of it too. I don't think we can merge this until/unless we have a full understanding of what the potential impact of this change will be.

sdillingham:
Because we're not in a position to merge and deploy this confidently, I'm going to close for now and we can reopen when we're ready to revisit this.

Issue #407, opened 2015-12-31
Separate Caching for Replacement and Static Items


ewr:
Currently, scprv4 uses memcache for caching. This works perfectly for the vast majority of SCPRv4 caching, where we're just stashing content that would otherwise have been computed / rendered inline. Old caches get expired as needed and a broken cache server just means we render those items in the request.

For items that are cached out-of-band, though, a missing server would mean missing content on the homepage. These items need to be cached into something other than memcache to remove the risk of a memcache restart taking out the homepage sections block.

I think it probably makes sense to create a `Cache` model and store this database into MySQL to give it persistence. From there it can be cached into memcache as normal, with only the periodic task responsible for expiring/updating the cache if it exists.

AC:
* Cache data is persistent and survives a memcache restart
* Cache data is persisted and cached out-of-band
* Frontend doesn't make any db calls for the cache items if they exist in memcache
* Frontend can fetch / cache the data from the persistent store if there is no memcache version or memcache is down

PR #427, opened 2016-02-08
Cache data in MySQL in case Memcached goes down

Ravenstine:
#407

I wrote a model that wraps around the Rails cache store that caches data in MySQL for items that are usually stored out-of-band.  This is so that, if a cache server goes down, we don't end up rendering all these items on every request.  If a cache server comes back online and there is an item that has been cached in MySQL, it will also write that data back to the cache server.

### DIFF
--- a/app/models/cache.rb
+++ b/app/models/cache.rb
@@ -9,12 +9,9 @@ class Cache < ActiveRecord::Base
       end
     end
     def write key, value
-      if return_value = Rails.cache.write(key, value)
-        return_value
-      else
-        create key: key, value: value
-        value
-      end
+      Rails.cache.write(key, value)
+      create key: key, value: value
+      value
     end
     def clear
       delete_all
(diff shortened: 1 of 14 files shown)


app/models/cache.rb
```diff
@@ -0,0 +1,20 @@
+class Cache < ActiveRecord::Base
+  class << self
+    def read key
+      if value = Rails.cache.read(key)
+        value
+      elsif value = where(key: key).limit(1).pluck(:value).pop
```
ewr: Can we change this to make `key` unique? That would just make your finder `find_by(key:key).try(:value)`

You'd just need to tweak the writer code to make it an upsert.

db/migrate/20160205232526_create_caches.rb
```diff
@@ -0,0 +1,8 @@
+class CreateCaches < ActiveRecord::Migration
+  def change
+    create_table :caches do |t|
+      t.string :key, index: true
```
ewr: You'll want to add a `unique:true` here if you make the change I suggested.

Ravenstine: :+1: :+1:

Issue #432, opened 2016-02-11
Allow episode rundowns to use any contentbase content


vsinghkpcc:
AC

- [x] Episode rundowns can use any contentbase content

Ravenstine:
@sdillingham So it wasn't too big a deal to get Rundowns to support other contentbase content, but I was foolish enough to not ask the question of whether or not this ticket implies displaying this support on Episode pages.  Is that something I should do?

sdillingham:
Yeah, we should have been more explicit with the AC. Once a rundown is able to contain anything from across contentbase, then you'll need to make sure they render  properly in a bunch of places. Off the top of my head:

* Episode Pages
* Program Landing Pages (which display the latest episode and everything in its rundown
* Segment Pages (which display the episode rundown after the segment body)
* Homepage Broadcast Bar (when the current program on air is a KPCC program where `is_segmented` is true, users can click on the Program Title and get a modal window with the episode rundown)
* The Program, Episode and Schedule endpoints on our API (both V2 and V3)

Ravenstine:
So I've worked displaying all rundown content on those pages, though I am not certain where we would be including them in the Program and Schedule endpoints, considering that we currently aren't returning segments or even episodes from those endpoints.  What might the rules be for delivering rundown content for those two endpoints?

sdillingham:
@Ravenstine you're right, the Schedule and Program endpoints aren't in play for this feature. My bad.

PR #443, opened 2016-02-22
Allow episode rundowns to include any content base content

Ravenstine:
#432 

Episodes can now be related to any Content Base model through rundowns.  So now, instead of simply having segments, an episode can have many segments, stories, shells, blogs, episodes, abstracts, events, pij_queries.  All of the content can be accessed at once using the `#content` method on the episode.

### DIFF
--- a/app/concerns/concern/controller/show_episodes.rb
+++ b/app/concerns/concern/controller/show_episodes.rb
@@ -3,7 +3,7 @@ module Concern
     module ShowEpisodes
       extend ActiveSupport::Concern
       def render_kpcc_episode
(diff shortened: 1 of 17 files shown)


Ravenstine:
@sdillingham Here is an example of a rundown with mixed content:
http://scprv4-staging.scprdev.org/programs/take-two/2016/02/09/12586/

There is an event, content shell, episode, and a segment in there.

The content is also reflected on the program page, since the rundown is featured:
http://scprv4-staging.scprdev.org/programs/take-two/

PR #445, opened 2016-02-23
Add a Rake task to make it easy to pull database backups

Ravenstine:
I decided to go ahead and create this.  It's a little dirty but I'd like to make it easier to pull backups.

Basically:

`rake db:pull (DEPLOYBOT TOKEN)`

and to start/stop the container:

`rake db:start` or `rake db:stop`

And to list both running and stopped containers:

`rake db:containers`

Merging this would mean requiring anyone doing SCPRv4 development to have Docker and Docker-Machine installed.  Theoretically, all one should have to do is run the `pull` task and be ready to go.


sdillingham:
Seems like a nice workflow improvement. :shipit:

PR #446, opened 2016-02-24
Make rundown content touch its parent episodes when it is updated

Ravenstine:
What I've done here is extracted the episode rundown relation code into a concern and included a before_save callback that will touch the parent episodes if the content has changed.  Theoretically, this should make it so that lists of episode rundown content in views can be fragment cached based on the episode object so that we do not always have to retrieve the individual pieces of content(which can come from multiple tables).


sdillingham:
Looks good @Ravenstine. Want to try merging and we can see what the transaction graphs look like in NR?

Issue #419, opened 2016-02-01
Improve Schedule Occurrence Conflict Detection Logic


Ravenstine:
![image](https://cloud.githubusercontent.com/assets/5193330/12727551/5ab14a36-c8d2-11e5-94b5-b0dcb8d196c7.png)

Sometimes, the ScheduleOccurrence#problems method will perceive a gap when an occurrence overlaps several occurrences, such as in the example above.

Also, since non-recurring occurrences are meant to take priority over other occurrences, overlaps involving them can probably be ignored.

PR #472, opened 2016-03-22
Schedule Occurrences should take large overlapping occurrences into account

Ravenstine:
#419 

The initial implementation of schedule occurrence problem detection had an issue with occurrences that would overlap multiple occurrences, including ones that were intended to overlap like special coverage.  

I decided to tread gaps and overlaps a bit different this time.  Overlaps iterate per-occurrence rather than per-pair(i.e. each_cons), and ignore non-recurring occurrences that overlap.  The preview in Outpost will show a list of occurrences that an occurrence is overlapping.  The occurrence it chooses is the larger one, or the earliest one if they are the same length, so we don't end up with duplicate listings.

The way that gaps are handled isn't too different from how they were previously, except the array of occurrences is "flattened" so that overlaps are eliminated and only gaps or a completely congruent schedule can be iterated over.


Ravenstine:
@sdillingham I'll keep what I've got on staging for a while if you want to play around and give any feedback.

sdillingham:
Hey @Ravenstine, I'm getting 500 errors on staging when I try to view conflicts on a recurring rule. Any thoughts about what the issue might be?

Ravenstine:
@sdillingham Give it another try now.  I don't know what changed after I deployed another branch earlier today and re-deployed this one.

sdillingham:
Thanks, it's working now. I'm noticing time zone discrepancies between what's listed in the schedule occurrences table view in Outpost and the Show Conflicts popup; not sure if that's a symptom of staging or not.

Issue #473, opened 2016-03-23
Add MoGo GTM - Remove old Pixel


vsinghkpcc:
AC

- [x] MoGo GTM code is added to scpr.org.
- [x] Old MoGo pixel is removed from scpr.org.

sdillingham:
Here's the spot in our code where we currently have a GTM tracking tag:

https://github.com/SCPR/SCPRv4/blob/master/app/views/shared/footer/_tracking_scripts.html.erb#L63-L69

sdillingham:
Also for reference (from the GTM docs):
https://support.google.com/tagassistant/answer/3207128?hl=en#multiple_tags

vsinghkpcc:
GTM code sent via Slack.

PR #478, opened 2016-03-25
Swap old Mogo pixel for GTM tracking script.

Ravenstine:
:point_up:
#473 

It's my understanding that having another GTM script should be just fine so long as it's using a different data layer.

### DIFF
--- a/app/views/shared/header/_document_header.html.erb
+++ b/app/views/shared/header/_document_header.html.erb
@@ -36,13 +36,7 @@
     }
   </script>
   <!-- mogo pixel -->
-  <noscript><iframe src="//www.googletagmanager.com/ns.html?id=GTM-KVHTMW"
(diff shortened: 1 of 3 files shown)


sdillingham:
Per the GTM docs, this tracking script should go in the <body> of our HTML documents, not in the <head>:

https://developers.google.com/tag-manager/quickstart

this partial is where the Mogo GTM script should go:

https://github.com/SCPR/SCPRv4/blob/master/app/views/shared/footer/_tracking_scripts.html.erb

This will keep all our tracking scripts in one place, rather than having separate partials for each tracking script spread out over the codebase. This partial gets inserted at the bottom of the <body> tag. While that runs contrary to Google's recommendation, we're making an opinionated decision in SCPRv4 to minimize latency in exchange for tracking codes loading a little later.

Issue #474, opened 2016-03-23
Set up our servers to run `npm install` on deployment


Ravenstine:
#459 

Now that we have a package management system for our external JS libraries, we need to figure out how we are going to deploy SCPRv4 with this change.  Right now, we don't have our servers set up to install/update vendor JS packages on each deploy so, while we can run locally and use CircleCI, pushing this work to staging and production fails.  This may mean digging in to Deploybot/Capistrano.  IDK yet.

AC: Figure out how we are triggering things like `bundle install` on deployment and add installment of NPM packages to that strategy.

Ravenstine:
@sdillingham 
There are a couple of options that I see so far.  The simplest would be to add a rake task to SCPRv4 that gets triggered before `deploy:updated`, which is also the same stage where Capistrano runs `bundle install`.  The advantage I see here is that we wouldn't have to touch Deploybot.

Alternatively, adding `require "capistrano/npm"` to Rails deploys in Deploybot would also work.  Though it would attempt this on every Rails deploy.  Not sure that's really a good thing.

I kind of lean towards the first solution.  What sounds good to you?

Ravenstine:
Actually, it also looks like we could add a cap file like this one:
https://github.com/capistrano/bundler/blob/master/lib/capistrano/tasks/hooks.cap

But a Rake file should also work, like this example:
https://github.com/capistrano/rails/blob/master/lib/capistrano/tasks/migrations.rake

sdillingham:
I agree that the first option is better; let's go with a rake task.

Ravenstine:
Seems I was mistaken about the first assumption; Capistrano runs tasks with before/after hooks if it's included inside a project, but Deploybot is including Capistrano as a library and it will currently only run the tasks that it is set up to:
https://github.com/SCPR/Deploybot/blob/master/lib/deploybot/run_deploy.rb

So I do think that, at some point, there will probably need to be some code included in the process that will check the project directory for `package.json`

Ravenstine:
Seems I was mistaken about the first assumption; Capistrano runs tasks with before/after hooks if it's included inside a project, but Deploybot is including Capistrano as a library and it will only run the tasks that it is set up to:
https://github.com/SCPR/Deploybot/blob/master/lib/deploybot/run_deploy.rb

So I do think that, at some point, the next preferred solution would be to include code in Deploybot that runs a custom task file in any Rails project that is deployed.  

Maybe something like this: 
```ruby
[...]

PR #485, opened 2016-04-05
NPM install Javascript packages on deployment

Ravenstine:
#474 

Most of our vendor JS assets have been removed from our repo and are now installed using `npm install` into the `node_modules` directory.  Browserify is also included, though is only currently used with moment-strftime(it is exported as a module rather than as a global).  The rest of our vendor javascripts are still required through Sprockets, but from the `npm_modules` directory.

A task has been added to `lib/tasks/deploybot.rake` which gets triggered by Deploybot on deployments.  This allows us to run `npm install`, and would allow us to run other project-specific tasks on deployment if we ever needed to.  Other Rails projects that aren't using NPM will still deploy normally.


Ravenstine:
@sdillingham I can see what you mean if we aren't using Handlebars in our actual codebase.  Do you still think it's acceptable to keep requiring it if we continue to use it with Outpost.Autosave?  Though it's only one instance of its use, I'd actually prefer Handlebars over Eco(which is being used by Outpost and is coming from it).  As much I adore CoffeeScript, Eco depends on CoffeeScript whilst Handlebars works with both, but is far more logic-less than Eco.

Does that make sense, or do you think we should still use Eco across the board?

sdillingham:
I think using/including Handlebars is fine for the autosave work.

Ravenstine:
Sweet.

Ravenstine:
@sdillingham Here's an update that I did on the packages I specified:

```
    fastclick           - 0.6.11         -> 0.6.11
    handlebars          - 1.0.0.beta.6   -> 1.0.2-beta
    history             - ???            -> 2.0.1
    jquery-ajax-xdomain - 1.0.4          -> 1.0.4
    jplayer             - 2.2.0          -> 2.8.4
    moment-strftime     - ???            -> 0.1.5
    moment-twitter      - ???            -> 0.2.0
    selectivizr         - 1.0.3b         -> 1.0.3
    swfobject           - 2.2            -> 2.2
    windowtamer         - ???            -> ???
    x2json              - 1.1.5          -> 2.2.0
    jquery-scrolldepth  - 0.9            -> 0.9.1
```

History and and moment-* had no version information.  Though I've confirmed that both the new versions of moment-strftime and moment-twitter work as expected.  Windowtamer never seems to have had a version number, but it was also last updated in 2012 which makes me doubt that the latest version is much different than what we already have.  Jquery-scrolldepth was minor bumped because that's been modified and forked for our own purposes.  Other version changes were the closest versions I could get to with NPM.

Issue #486, opened 2016-04-07
Autosave in Outpost: Finish MVP


vsinghkpcc:
AC

- [x] Include related content and assets in the autosave
- [x] Remove the timestamp from the save bar
- [x] Revise modal format and message
- [x] There is an alert modal when a user tries to close an edit screen and has unsaved changes

Notes: Vijay will provide messaging for both modals.

PR #487, opened 2016-04-14
Improve the reliability of autosave in Outpost

Ravenstine:
This includes work for #486 and #485, which means that this includes autosave work and the work to deploy with vendor packages using NPM.

There's now a script in `app/assets/javascripts/outpost/autosave.js.coffee` that tracks fields, Backbone collections, and some elements and stores those in a local snapshot using PouchDB.  If the autosave script detects a difference between the snapshot and the fields loaded on the page, it gives the CMS user the option to restore from the snapshot or dismiss it.  Nearly every field gets saved with the exception of audio, which gets a little more complicated but can be included in a future task.

A warning window is also included that, if changes have been made to the page, will ask them whether or not they really want to navigate away from the page.  This should help prevent data-loss in most cases, but the autosave will still be there in case of connection failure, dead battery, etc.

I would consider merging this in to Outpost itself when this has matured and proved successful in production.

NOTE: The change to `app/models/article.rb` is to prevent it from failing when the database includes a column that it isn't expecting.  I can't think of a valid reason why we would want that to happen.


Ravenstine:
@sdillingham This is on staging currently, so feel free to make any comments QA-wise.

PR #509, opened 2016-04-18
Add noindex meta tag to staging to prevent Google from crawling it

Ravenstine:
:point_up:


sdillingham:
Looks good @Ravenstine. :shipit:

PR #515, opened 2016-04-26
Outpost Autosave Release 2

Ravenstine:
We had to roll back the last release because the keyup/down events we were tracking were conflicting with key combos.  Instead, we're now shimming DOM objects so that we can directly track changes on all of them, including things like the textareas that are being dynamically modified so their change/input events aren't triggered.

### DIFF
--- a/app/assets/javascripts/outpost/autosave.js.coffee
+++ b/app/assets/javascripts/outpost/autosave.js.coffee
@@ -128,7 +128,7 @@ class outpost.Autosave
     query = []
     query.push("#main #{elName}[id]") for elName in @elementNames
     query = query.join(", ")
-    $(query).not(':button').not('[type=hidden]').not('#autosave-revisions')
+    $(query).not(':button').not(':hidden').not('#autosave-revisions')
 
   # private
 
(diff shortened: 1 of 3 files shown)


sdillingham:
@Ravenstine: Some edge case bugs I found that we should address before this is merged to master and deployed:

## Asset Chooser

STR (I performed this test using a PIJ Query record):
1) Pop up asset chooser on a story
2) Select one new asset and click "save and close"
3) Pop up asset chooser again
4) Delete the asset you just added, and click "save and close"
5) Click "Save" on the story in Outpost.
6) When the record saves and the page reloads, you should see an Autosave Restore modal pop up.
7) Click "restore changes" and you should see that the deleted asset has reappeared on the story.

## Related Content

STR (Also tested on a PIJ query):
1) Scroll to a "related content" aggregator on a story edit view
2) Click on the recent tab to the right of the aggregator collection
3) drag a story from the Recent list into the aggregator collection
4) drag the story out of the aggregator collection to remove it
5) Wait a few seconds and then close the current browser tab
6) reopen the tab
7) you should see an Autosave Restore modal pop up.
8) Click "restore changes" and you should see that the removed related content has reappeared in the aggregator.

PR #587, opened 2016-07-19
#get_article should be called instead of #to_article so that we aren't generating an article when we already have one indexed.

Ravenstine:
Looking at staging, I'm not really sure how much of a temporary improvement this will be, but it should be safe to deploy and that might provide a better indication.


Ravenstine:
#588

Issue #612, opened 2016-08-12
Double articles being pulled into remote articles


ezass:
Looks like we're getting double posts from NPR. FYI. 
![screen shot 2016-08-12 at 11 00 58 am](https://cloud.githubusercontent.com/assets/994636/17632389/ed79d9f4-607c-11e6-898e-87815344e110.png)

Ravenstine:
@ezass I am pretty confident this has been fixed since I have merged a change into production, and no new duplicates occur no matter how many times I press the sync button.  Ping me if the problem recurs.

Ravenstine:
Nevermind, the problem seems to have recurred.  I may have to prevent duplicates at the database level or something along those lines.  It all seemed to be working as expected yesterday.

Ravenstine:
Perhaps it's related to a race condition:
https://stackoverflow.com/questions/26500908/concurrent-cause-duplicate-records-inserted

Just a guess, since we seem to run into this problem on occasion in completely different parts of the code without good or any reasons for why it would happen at the application level.

Ravenstine:
Ok, we are definitely no longer seeing duplicates.  Remote articles now have uniqueness set to true on one of its indexes.  I'd still like to figure out where there are two competing processes trying to download remote articles, but I'd eat my hat if the database level constraint doesn't work.

PR #613, opened 2016-08-15
Prevent NPR remote articles from being pulled in multiple times

Ravenstine:
#612 

I have a suspicion that the duplication was being caused by, at some point, the article_id being cast as a different type from the fetched article ID(string being evaluated against an integer).  It's not clear to me why this didn't seem to always happen, but the article_id, when queried from the database, comes back as a string.

So I did two things:

- Convert both the ID in the database and the fetched ID to strings during comparison.
- Fetch existing IDs from the database in bulk as opposed to calling an #exists? query on every article fetched.  This makes the importer much faster.

On staging, we were seeing up to 3 duplicates per story.  After implementing this change, no duplicates come in no matter how many times I run the importer, so I think this is worth deploying to production to see if this issue can be resolved today.

### DIFF
--- a/app/importers/npr_article_importer.rb
+++ b/app/importers/npr_article_importer.rb
@@ -36,8 +36,16 @@ module NprArticleImporter
 
       added = []
 
-      npr_stories.reject { |s|
-        RemoteArticle.exists?(source: SOURCE, article_id: s.id)
+      fetched_story_ids    = npr_stories.map(&:id)
+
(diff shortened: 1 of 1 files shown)


Ravenstine:
It's possible that the IDs were always comparing a string against an integer, and the timing of the importer and the publish_at of the stories in the NPR API coincidentally prevented most of the duplicates.  But that's speculation without much evidence.

Issue #665, opened 2016-10-19
Application Errors on SCPRv4 should report to something


nakedsushi:
It doesn't look like we're logging errors anywhere and New Relic's error reporting is not detailed enough for us to the answer the question, "Why did X give an error?"

Install some sort of error reporting, either Airbrake ($49/mo) or Honey Badger (Free for non-commercial open-source) in the SCPRv4 codebase so that we can figure out what's breaking.

AC:

- [x] an error in SCPRv4 notifies either Honey Badger or Airbrake
- [x] an error in any resque worker notifies Honey Badger or Airbrake

nakedsushi:
I implemented Honeybadger and have deployed the changes on staging and production.  I triggered an error on staging and it reported to Honeybadger.  Next time we see an error in production or a background worker, we should see it in Honeybadger.

PR #671, opened 2016-10-24
Adds Honeybadger gem to report errors

nakedsushi:
@Ravenstine Can you take a look at this? Let's pair tomorrow morning so you can show me how to deploy to staging and then production.

https://github.com/SCPR/SCPRv4/issues/665


Ravenstine:
:shipit:

PR #788, opened 2017-03-14
exit intent modal

arellaTV:
@nakedsushi 

General notes:
- Implemented via exit-modal.js in assets/javascripts folder
- Imported this script via the shared.js in the same folder
- Created a custom stylesheet exit-modal.css in assets/stylesheets folder
- Imported the stylesheet to multiple application.css.scss files to include it sitewide
- Uses jQuery for DOM selection, creation, and event listening/binding

Behavior:
- Modal opens on mouse leaving the top viewport for desktop and quickly scrolling up for mobile
- Modal stays closed after dismissing it once, until the page refreshes
- Modal can be closed via clicking out of the modal or clicking the 'x'
- Modal functionality exists in every page of the application currently


app/assets/javascripts/exit-modal.js
```diff
@@ -0,0 +1,81 @@
...
+
+  // Checks if the current device is mobile
+  var isMobile = false;
+  if(/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)) {
+    isMobile = true;
+  }
+
+  // Builds individual parts of the modal
+  var body = $('body');
+  var modal_container =$('<div id="modal-container"></div>');
```
nakedsushi: Our CSS naming convention for CSS classes is to follow BEM (http://getbem.com/introduction/) when we can and to be as descriptive in as few characters as we can, if that makes sense. For example, for the exit modal container's id, instead of the id being named `modal-container` perhaps try `exit-modal`.

Some examples of our take on BEM naming can be found in app/assets/stylesheets/better_homepage/application.css.sass

For javascript variables and function names, we try to adhere to camelCase so instead of `var modal_container =...` try `var modalContainer = ...`

app/assets/javascripts/exit-modal.js
```diff
@@ -0,0 +1,81 @@
...
+  var underlay = $('<div class="underlay"></div>');
```
nakedsushi: Following the above BEM's convention, the class `underlay` can become `exit-modal__underlay`

app/assets/javascripts/exit-modal.js
```diff
@@ -0,0 +1,81 @@
...
+  // Assembles the modal from the individual parts and attaches to body
+  modal.append(close_button);
+  modal_container.append(underlay);
+  modal_container.append(modal);
+  body.append(modal_container);
+
+  // Functions to open and close modal
+  var open_modal = function() {
+    if (!modalVisited) {
+      modal_container.css('display', 'block');
```
nakedsushi: Instead of having the javascript modify the style directly, consider using javascript to change the CSS class of the object.  In this example, instead of changing modal_container's `display` attribute to `block`, consider having a separate CSS class in the stylesheets like `exit-modal__underlay--shown` and `exit-modal__underlay--hidden`

app/assets/javascripts/exit-modal.js
```diff
@@ -0,0 +1,81 @@
...
+      modalVisited = true;
+    }
+  }
+  var close_modal = function() {
+    modal_container.css('display', 'none');
+  }
+
+  // Listens for when user is about to exit the page
+  if (isMobile) {
+    $(window).on('scroll', function(e) {
```
nakedsushi: Let's extract this anonymous function into a named function.  Wondering if we should throttle this so that it doesn't fire several times a second, which can severely slow down a mobile browser? https://css-tricks.com/debouncing-throttling-explained-examples/

app/assets/javascripts/exit-modal.js
```diff
@@ -0,0 +1,81 @@
...
+  underlay.on('click', close_modal);
+  close_button.on('click', close_modal);
+  $(window).on('keyup', function(e) {
+    if (e.keyCode == 27) {
+       close_modal();
+    }
+  });
+
+  // Helper function that checks scroll speed
+  var checkScrollSpeed = (function() {
```
nakedsushi: Oh wait, did you already add a throttler here? If so then ignore my comment above about throttling.

app/assets/stylesheets/exit-modal.css
```diff
@@ -0,0 +1,41 @@
+#modal-container {
```
nakedsushi: Let's do SASS here instead of CSS.  We can be pretty general in this case, so let's leave `font-family` off until we really want to use a specific font. We'll just leave that to the parent CSS to take care of for now.

app/assets/stylesheets/exit-modal.css
```diff
@@ -0,0 +1,41 @@
...
+  z-index: 3;
+}
+
+#modal-container .underlay {
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  top: 0;
+  left: 0;
+  background-color: rgba(0,0,0,0.5);
```
nakedsushi: Do we have an existing SASS variable for color? Something like `$color-gray-light` or one of those?

Ravenstine: @nakedsushi It depends on if the [scpr-style-guide](https://github.com/scpr/scpr-style-guide) is being imported on the page(which I think it is).  If it is, all the color variables can be found [here](https://github.com/SCPR/scpr-style-guide/blob/master/src/css/util/_vars.scss).

app/assets/javascripts/exit-modal.js.coffee
```diff
@@ -0,0 +1,70 @@
...
+    $(window).on('scroll', throttledScrollHandler)
+  else
+    $(window).on('mouseleave', mouseLeaveHandler)
+
+  # Listens for when user closes the modal
+  underlay.on('click', closeModal)
+  closeButton.on('click', closeModal)
+  $(window).on('keyup', closeModal)
+
+  undefined
```
nakedsushi: nitpick: if we definitely do not want to return anything, just do `return` here instead of `undefined`.

app/assets/stylesheets/exit-modal.css.scss
```diff
@@ -0,0 +1,48 @@
+#exit-modal {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 3;
+}
+
+#exit-modal .exit-modal__underlay {
```
nakedsushi: Thanks to BEM, we don't need the first part `#exit-modal` here or in any of the other classes below. It's already pretty clear that `.exit-modal__underlay` will select the correct object.  If this is to ensure that `.exit-modal__underlay` doesn't apply a style to an incorrect object, then we should cross that bridge when we get there.

app/assets/javascripts/exit-modal.js.coffee
```diff
@@ -0,0 +1,81 @@
...
+    $(window).on('scroll', throttledScrollHandler)
+  else
+    $(window).on('mouseleave', this.mouseLeaveHandler)
+
+Modal.prototype.bindCloseEventListeners = ->
+  $('.exit-modal__underlay').on('click', this.closeModal)
+  $('.exit-modal__close-button').on('click', this.closeModal)
+  $(document).on('keyup', this.closeModal)
+
+Modal.prototype.capturePageInfoFromURL = ->
```
nakedsushi: Instead of basing the modalContent on the URL, consider using Rails to your advantage. Just like the way you set `@is_pledge_time`, you can set `@exit_modal_type` to some string in your controller, instantiate the javascript variable in the same script tag that you're doing @is_pledge_time in the view, and base the modalContent off of that.

My concern with using the URL is that it may be less robust if the URL ends up changing, which can happen if we end up changing our rails routing.

arellaTV: I thought the same thing! I'm actually working on that now :)

arellaTV:
@nakedsushi 
Modal now only displays on show segments, and script is no longer shared site-wide

nakedsushi:
@arellaTV changes look good. Can't wait to see it all tied together with the 3rd issue.

arellaTV:
@nakedsushi 

- 'Which show' logic is transferred from javascript to rails controller
- When AdminUser creates a flatpage with a path, 'exit-modal/{show}', the html is applied to the exit modal's contents
- When no flatpage is included, no exit-modal is created
- Exit-modal's width is responsive to window (according to three breakpoints)
- Removed DOM creation logic from javascript and instead declare it in segment html template

Updates incoming!

app/controllers/programs_controller.rb
```diff
@@ -97,6 +97,19 @@ def segment
     @program = @kpcc_program = @segment.show
     @featured_programs = KpccProgram.where.not(id: @program.id, is_featured: false).first(4)
 
+    @is_pledge_time = PledgeDrive.happening?
+    @pledge_time_flatpage =  Flatpage.visible.find_by path: 'exit-modal/pledge-drive'
+    @program_flatpage = Flatpage.visible.find_by path: 'exit-modal/' + @program.title
+    @default_flatpage = Flatpage.visible.find_by path: 'exit-modal/default'
```
nakedsushi: I like that you extracted the variable into @modal below so that the view only has to care about one variable.

Since these 4 variables are only use in the controller (and not in the view), let's make them into local variables (without the @ sign) instead.

In general, we only use instance variables (the ones with @) if we're going to refer to them in the view.

arellaTV: @nakedsushi 

Okay! On board for a demo when you're ready :)

nakedsushi:
I think this got merged, so closing it.

PR #793, opened 2017-03-23
Fix: Audio player has context parameter on show landing page

arellaTV:
- Uses helper function `url_with_params` to pass in parameters
- Uses the instance variable, `@program.slug`, to supply show's context value
- Also adds 'via: website' parameter, similar to the episode page
- Network tab shows the parameter `context=offramp` in both Off-Ramp's landing page and episode page


app/views/programs/kpcc/_featured_story.html.erb
```diff
@@ -17,7 +17,8 @@
           <mark class="play">
             <ul>
               <li class='story-audio'>
-                <a href="<%= featured_story.audio.first.url %>" class="play-btn audio-toggler" title="<%= featured_story.short_title %>" data-duration="<%= featured_story.audio.first.duration %>">
+                <% url = url_with_params(featured_story.audio.first.url, context: @program.slug, via: 'website') %>
+                <a href="<%= url %>" class="play-btn audio-toggler" title="<%= featured_story.short_title %>" data-duration="<%= featured_story.audio.first.duration %>">
```
nakedsushi: Since this is the only place that's using the `url` variable, what do you think about just ditching it and directly calling `url_with_params()` here so that it's:
`<a href="<%= url_with_params(featured...) %>" class=...`

arellaTV:
- Had to resolve some dependency/ circle issues. Involved:
  - Adding new IP's to allow in Webmock
  - Reverting to previously working Gemfile
  - Nilchecking strip_inline_assets
    - Returning the original body

PR #794, opened 2017-03-23
Fix: Cutoff ads in Windows Firefox

arellaTV:
- Sets `position` property of google's ad container to `relative` to offset weird Windows/Firefox behavior that restricts setting dimension properties to children of an `absolute` positioned div.
  - Wrote another helper function called `setPositionRelative` in the `adSizer` class in `utilities.js.coffee`. It uses jQuery to find #google_image_div and sets its `position` to `relative`.


nakedsushi:
:shipit:

nakedsushi:
Merged into master here already: https://github.com/SCPR/SCPRv4/commit/3b9c4bd0a5b5a280a3587e8023559c1a8a63a1a2

PR #807, opened 2017-04-06
Renders video meta tag to articles with video content

arellaTV:
- The icon lives in `public/static/images/icon_line-video.svg`
- Adding the "Video" string to the array concatenates to the path to the svg


nakedsushi:
Looks good to me. Go ahead and merge it, @arellaTV

PR #814, opened 2017-04-12
Podcast list

arellaTV:
@Ravenstine 
- Made two cells unique to this view:
  - `podcast_list` cell
  - `popular_articles` cell (amended with side_bar.erb)
- Following convention:
    - Tried to keep nesting at a minimum
    - Classes and ids follow BEM naming convention, `o-podcast-list__description`
- Layout styling is found in `app/assets/stylesheets/layouts/podcasts.css.sass`
  - Uses `blogs.css.scss` as a starting point
- The podcast view itself is now only concerned with importing the cells
- The view reuses a lot of cells that were already defined

Amending popular_articles
- I wasn't sure if the popular articles side bar template should be its own cell or amended to the already existing popular_articles cell. For now I choose the latter because the information is relevant between both.
  - I store the most viewed and commented articles from the cache into instance variables `@popular_articles_viewed` and `@popular_articles_discussed` respectively. And then I render them via the `side_bar` method. Does this look okay?
https://github.com/SCPR/SCPRv4/compare/unification...podcast-list?expand=1#diff-e9d2884888b2e70f84bdda060f9bbfb8

- My exit modal code is in this repo from merging unification with master. Should I try to pull from the `unification` branch again, and apply the new changes to that new branch?


Ravenstine:
Checked it out, and it looks good to me.  As with the previous pages I worked on, we'll need to address responsiveness later on.  For now, it's most important that the pages use the new design, and the construction of the markup and the styling is consistent with the rest of the pages that have been developed so far. 👍

PR #857, opened 2017-05-12
Renders related links cwidget in segment pages

arellaTV:
## Related Links CWidget
- The `content` object being fed into `/views/shared/cwidgets/_related_links.html.erb` was resolving with empty array values for `content.links` and `content.related_content`
- I changed the way we access these properties in the view so that we're passing in `article` now instead of `article.get_article`. I also removed the `links.present?` check since Rails was telling me that there was no `links` property.
- The change should only affect the `segment` view specifically.

Imgur


arellaTV:
Changed files can be found here:
https://github.com/SCPR/SCPRv4/pull/857/files

arellaTV:
@nakedsushi 

Okay! I think I found a slightly better solution that is less intrusive and only changes how the related content is being accessed. Rather than remove the `article.get_article` from `programs/kpcc/_segment.html` and `content.links.present?` from `shared/cwidgets/_related_links.html`, it might be better to change `content.related_content.present?` to `content.original_object.related_content.present?`.

Unfortunately still doesn't address the infinite loop, but doesn't effect anything else besides the cwidget. Also, if there is content from related_links, those will also show.

arellaTV:
Pushed to production and confirmed to work!

PR #882, opened 2017-06-30
Changed link for "Update Checking Account" in navigation menu

arellaTV:


nakedsushi:
:shipit:

arellaTV:
@nakedsushi, @Ravenstine 
- I locked down the npm version in circle.yml to 5.1.0 which seems to have fixed the npm install failures (there was a breaking bug with the pacote dependency that is fixed in this version).
- In outpost, I removed a duplicate slug form from the edit events page introduced here: https://github.com/SCPR/SCPRv4/compare/ea2ebdc7f6db...558841fc929c, which should fix the event tests that are failing.

Ravenstine:
LGTM 👍

Issue #919, opened 2017-08-04
Outpost: CKEditor strips <form> tags on outpost refresh


arellaTV:
For some reason, CKEditor strips out `<form>` tags on refresh. For example, let's say you build a flatpage with a form inside of it. When you come back to edit that flatpage again in a new browser window, and you change the title of the flatpage and click save. The resulting flatpage will have its `<form>` tags missing.

AC:
- [ ] form tags persist through refresh

PR #924, opened 2017-08-09
Removed form specific code from the program controller

arellaTV:
I think it makes more sense to give the `segment` method of the `program` controller less responsibility in terms of what forms should appear in the modal. By removing the form logic from the `program` controller and `segment` view, you can essentially replace the forms on the fly within the flatpages.

I avoided pasting the forms into the flatpages the previous time I worked on this because CKEditor would strip the form out on refresh, but I think since the flatpage itself isn't being edited very frequently, I think it's probably safe to just let the flatpage deal with the form. In the meantime, I've made a ticket to try and solve CKEditor's form stripping bug once and for all (https://github.com/SCPR/SCPRv4/issues/919)

EDIT (2017-07-09 2PM): I also silenced the test spec that tests the Eloqua job #async_send_email


spec/concerns/concern/methods/eloqua_sendable_spec.rb
```diff
@@ -66,16 +66,16 @@
...
+      #   Job::SendEmailNotification, "TestClass::Alert", 999)
 
       alert.async_send_email
     end
   end
 
   describe "#eloqua_config" do
     phoney_class = Class.new do
-      include Concern::Methods::EloquaSendable 
+      include Concern::Methods::EloquaSendable
```
nakedsushi: Let's just go ahead and git rid of the whole `describe "#eloqua_config" do` block completely.

arellaTV:
Rebased to one commit and pushed to master: 526a14df440e713a8e27319d6444609e317a4604
PR is now shipped to production! 

Here's an example of a page with a modal that has a form in it:
http://www.scpr.org/programs/the-frame/2017/08/08/58464/aubrey-plaza-basks-in-the-discomfort-of-ingrid-goe/

Issue #927, opened 2017-08-10
Update embeditor-rails for https


nakedsushi:
the api URL is hard coded into the gem

Update our version of embeditor-rails so that it replaces `http` with `https` in this file: https://github.com/SCPR/embeditor-rails/blob/05a1f8ef4d3cddc02b0e02b478a574d7c189ac51/app/assets/javascripts/embeditor/adapters/embedly.js.coffee

AC:
- [ ] embeds work on https://www.scpr.org now because it's not mixed content

nakedsushi:
Pending review:
https://github.com/SCPR/embeditor-rails/pull/7

nakedsushi:
Updated Gemfile in scprv4 to point to new version of gem:
https://github.com/SCPR/SCPRv4/pull/930

nakedsushi:
Shipped to production: https://github.com/SCPR/SCPRv4/commit/75c38fd4343ad7763ed8db442d6c33497a563915

PR #930, opened 2017-08-11
Uses updated for https version of embeditor-rails

nakedsushi:
Updating the Gemfile since I updated the gem.

https://github.com/SCPR/SCPRv4/issues/927


arellaTV:
Makes sense. Looks like it's good to go!

PR #931, opened 2017-08-11
Add fields to programs api

Ravenstine:
Adds `description_text` and `phone_number` to KPCC programs and external programs in the API and Outpost.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,6 +70,7 @@ services:
     ports:
       - 9200:9200/tcp
       - 9300:9300/tcp
+      - 9250:9250/tcp
     volumes:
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -70,6 +70,7 @@ services:
     ports:
       - 9200:9200/tcp
       - 9300:9300/tcp
+      - 9250:9250/tcp
```
nakedsushi: What does this do?

Ravenstine: It's apparently needed to initiate a "cluster" during testing.

nakedsushi:
:shipit:

Issue #921, opened 2017-08-08
Automated Newsletters solution to support transition to SFMC Tweaks


vsinghkpcc:
AC

- [x] For The Short List, the automated output uses the new SFMC Live Intent Ad Tags. (ask Vijay for the Tags)
- [x] Strip out any UTM code included in the HTML output (for both the Short List and Breaking News Alerts)

vsinghkpcc:
- [ ] For the Short List Editions HTML preview output, replace 300x50 ad code table on line ~58 with:
`<table border="0" cellpadding="0" cellspacing="0" ><tr><td colspan="2"><a style="display: block; width: 300px; height: 50px;" href="http://li.scpr.org/click?s=216556&sz=300x50&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" rel="nofollow"><img src="http://li.scpr.org/imp?s=216556&sz=300x50&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" border="0" width="300" height="50"/></a></td></tr><tr style="display:block; height:1px; line-height:1px;"><td><img src="http://li.scpr.org/imp?s=216557&sz=1x1&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" height="1" width="10" /></td><td><img src="http://li.scpr.org/imp?s=216558&sz=1x1&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" height="1" width="10" /></td></tr><tr><td align="left"><a href="http://li.scpr.org/click?s=216544&sz=116x15&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" rel="nofollow"><img src="http://li.scpr.org/imp?s=216544&sz=116x15&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" border="0"/></a></td><td align="right"><a href="http://li.scpr.org/click?s=216545&sz=69x15&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" rel="nofollow"><img src="http://li.scpr.org/imp?s=216545&sz=69x15&li={LIST_ID}&e=%%emailaddr%%&p=%%jobid%%" border="0"/></a></td></tr></table>`

- [ ] Replace 300x250 ad code table on line ~183 with:
[...]

PR #934, opened 2017-08-14
Automated newsletter changes

nakedsushi:
Updates to the template that generates HTML for the Short List and Breaking News, per Vijay's specs: https://github.com/SCPR/SCPRv4/issues/921


app/models/edition.rb
```diff
@@ -118,9 +118,6 @@ def sister_editions
   end
 
   def html_body
-    # NOTE: This is ugly as hell but will only be around for the short term.
```
Ravenstine: Is there some objection to the word "hell"?  We shouldn't have those methods hanging around unless they are replaced with something besides a call to EloquaEmail.  The model shouldn't even be rendering HTML; this was meant to be a stopgap because Eloqua was going away very soon.  Both Editions and EloquaEmail models need serious cleanup.

nakedsushi: No, the objection is to having a vague comment. If it's temporary code, it should link to an issue with more context, or a TODO with instructions to fix it.

Ravenstine: Gotcha.  Well, that same comment occurs a few times beneath(there are multiple methods that shouldn't exist in that model), so I will make a commit to revise them to have more clarity.

nakedsushi:
merged here: https://github.com/SCPR/SCPRv4/commit/01aded68e1500ff082e0f46aea3da6f6479fd841

PR #936, opened 2017-08-16
Schedule feedback

arellaTV:
**Staging URL:** https://scprv4-staging.scprdev.org/schedule
**Katie's mockup:** https://github.com/SCPR/design/issues/212

**Changelog:**
- Schedule page now shows a daily schedule instead of a weekly one (as per @tinykb's feedback)
- Next and Previous buttons are included in the header (in the form of date days, e.g. Mon., Wed.)
- Single days can be found using path parameters (e.g. schedule/2017/08/16).

**Suggested review path:**
- The best place to start is `spec/controllers/programs_controller_spec.rb`
  - It exposes the `@schedule_occurence` instance variable to the view (`app/views/programs/schedule.html.erb
`) and feeds it into the cell as `date` (`app/cells/schedule/show.erb`).
- The cell controller (`app/cells/schedule_cell.rb`) is responsible for formatting the date into days (e.g. Mon.).

**These files are style changes or updates in markup:**
 - `app/cells/regular_features_list/show.erb`
- `app/cells/schedule/show.erb`
- ` app/cells/schedule/style.sass`

**Testing:**
I also wrote update the `GET /schedule` tests to reflect the above changes, and I added a new one to test whether a schedule occurrence shows up when given path params.


app/controllers/programs_controller.rb
```diff
@@ -127,10 +127,17 @@ def archive
   end
 
   def schedule
-    @schedule_occurrences = ScheduleOccurrence.block(
-      Time.zone.now.beginning_of_week, 1.week, true
-    )
+    @date = Time.zone.now.beginning_of_day
+
+    if params[:year] and params[:month] and params[:day]
```
nakedsushi: Consider some defensive programming here. What happens if any of these params are strings like 'asdf' instead of expected numbers?  Time.zone.local might raise an error and crash the page.

app/cells/schedule/style.sass
```diff
@@ -7,12 +7,17 @@
...
-  padding: 1em
-  text-transform: uppercase
-  font-weight: bold
-  color: $color-gray
+  padding-top: 1em
+  padding-bottom: 1em
+  display: flex
+  justify-content: space-between
+  align-items: center
+  .o-schedule__list-date-header__current-date
```
nakedsushi: Is there supposed to be a newline here?

app/controllers/programs_controller.rb
nakedsushi: Weird, it reverted to my old comment.  What I meant to say is, is the change in `routes.rb` sufficient to make sure people passing in 'asdf' as one of the params (month, year, or day) won't make Time.zone.local barf?

arellaTV:
**To recap the changes:**
- `/schedule/asdf/08/13` will yield a root path error, redirecting to a 404 page
- `/schedule/2017/08/42` will default to today's schedule
  - the controller validates the params first to see if it can yield a proper date
  - in the above instance, the day is out of range, so the validation function returns false

PR #959, opened 2017-09-19
Fixes missing audio element from news story amp pages

arellaTV:
- This is basically a copy from the audio element found in `views/amp/audio.html.erb`, a template that is served for KPCCProgram Segment pages.
- NewsStory Article uses the `views/amp/default.html.erb` template and should also have this but currently doesn't.
- Falls back to nothing if no audio is found.


app/views/amp/default.html.erb
```diff
@@ -20,6 +20,17 @@
         <%= @amp_record.asset.caption %> <span class="credit"><%= @amp_record.asset.owner %></span>
       </figcaption>
     </figure>
+    <% if audio = @amp_record.audio.first %>
+
+      <% url = url_with_params(audio.url, via: 'website', context: @amp_record.to_article.context) %>
```
nakedsushi: Will we always have audio.url? If it's a non-null field, fine to leave as is. If not, we should add a nil check in case url is nil?

arellaTV:
@nakedsushi 

**Latest changes:**
- Segment pages now point to the `amp/segment` template instead of `amp/audio`
- `amp/audio` is revised to only be the audio component and used as a partial
- in both `default.html.erb` and `segment.html.erb`, the audio parital is only rendered if the audio record is `present?`

PR #963, opened 2017-09-21
Article responsiveness

arellaTV:
Staging URL: https://scprv4-staging.scprdev.org/news/2016/06/01/61067/voter-turnout-for-presidential-primaries-is-usuall/

This is a quick PR to see if I'm on the right track in terms of the mobile article pages.
Major changes in this current commit:
- Font sizes adhere to type scale conversions
- Ordering of cells matches the ordering in SCPRv3
- Spacing between cells span two rows (60px), unless there's a horizontal rule (which already has 30 pixels above and below it).
- Spacing inside cells stay relatively the same, or at least adhere to the multiple of 15px rule


tinykb:
This is looking great! Only a couple lil notes: 
1 - Charts are getting all squished
2 - Last graf is bigger than the rest?
3 - Sign up button is still set in Fira instead of FranklinGothic
4 - Bullet points are hovering a little above the x-height line
5 - Footer can be condensed (see Footer - Small.jpg)
[Article - Mobile.zip](https://github.com/SCPR/SCPRv4/files/1322142/Article.-.Mobile.zip)

PR #966, opened 2017-10-03
Adds div plugin to CKEditor and fixes save issues with div els

arellaTV:
Changelog:

- Fixes an issue where inline-styles added to `<div>` elements are stripped out of the body after refreshing by loosening the allowedContent rules in CKEditor.
- Adds the `Div` plugin, which allows an easier method for grouping of elements and applying styles or classes to them.
- Adds the classnames, `.group-wrapper`, `.group-right`, and `.group-left` for a convenient way to float groups right, left, or cleared on both directions.


app/assets/javascripts/outpost/ckeditor/custom-config.js
```diff
@@ -18,7 +18,7 @@ $(function() {
       "i",
       "b",
       "blockquote",
-      "div[*]",
+      "div[*](*){*}",
```
nakedsushi: What does the (*){*} do?

arellaTV: @nakedsushi 

I believe the `(*)` denotes that you can apply any class to any `div` element. `{*}` allows for any inline style. Is this maybe too loose? An alternative could be:

`div[*](group-wrapper, group-right, group-left){ float }`

arellaTV:
@nakedsushi 
This is mainly a way to address edge cases in which images that are floated to the right (usually by default because articles float portrait images to the right) overlap with other elements below it. This adds a button so you don't have to add a div and class manually in the source window.

arellaTV:
Got it! Thanks @nakedsushi!

PR #974, opened 2017-10-07
List pages responsiveness

arellaTV:
Some notes:
- Schedule page was a bit tricky in tablet mode. I decided to just show the daily schedule and regular features section side by side with a smaller margin in between them. If it makes more sense for the regular features to wrap to the bottom of the daily schedule, I can do that instead.
- Search Results page was a bit tricky, but I was able to add the "PROGRAMS" feature along the top of the search results. It isn't laid out exactly like the mockups, but it's the closest I could get despite the google result limitations

Links:
- Schedule
https://scprv4-staging.scprdev.org/schedule
- Program List
https://scprv4-staging.scprdev.org/programs
- Podcasts
https://scprv4-staging.scprdev.org/podcasts
- Search Results
https://scprv4-staging.scprdev.org/search/?q=air
- Blog Directory
https://scprv4-staging.scprdev.org/blogs


tinykb:
**Schedule** – The schedule isn't working for me on tablet/middle-sized windows
screen shot 2017-10-10 at 12 36 34 pm


**Podcasts** – This page doesn't respond past tablet size
screen shot 2017-10-10 at 12 38 09 pm


**All these pages** – Is Tiny Rick in play here? I'm not seeing the fonts scale down, and they're looking pretty large when we get to small tablet/mobile

arellaTV:
Oops! Staging was on the wrong branch. Newest list pages are back up!

tinykb:
Okay, my only notes now are: 
1) The podcast tiles get *really large* when /podcasts is on smaller screens. Not sure if getting rid of the artwork altogether is the way to go here, but it might be.
2) Wondering if we should ditch show artwork when we get down to small screens in search results, since they get little/kind of awkward

PR #976, opened 2017-10-14
Adds the ability to edit the contents of the KPCC In Person Page via Outpost

arellaTV:
I'm so sorry for the big PR!

Specific files to look at:
- The migrations
- Models:
  - `landing_page`
  - `landing_page_events`
  - `landing_page_reporters`
- The events controller

The Landing Page model is heavily inspired by the Vertical model, except the associations are specifically for events and reporters at the moment. The biggest reason that this is a separate model is due to the fact that landing pages will probably have different needs as compared to verticals down the line (won't necessarily need a featured quote, featured interactive, tags, etc). Also, I wanted Verticals to stay focused on being "super-categories" if you will as opposed to being the home for all landing pages.

It uses the outpost aggregator to control the content and order of the top 3 featured events at the top of the events landing page. The slug is mainly used here to allow for the "View Live" button at the bottom of the edit page.

Outpost Edit Page: https://scprv4-staging.scprdev.org/outpost/landing_pages/3/edit
KPCC In Person Landing Page: https://scprv4-staging.scprdev.org/events/kpcc-in-person

One caveat is that I had to go into rails console in staging and add a `Permission` record manually:
`Permission.create(resource: 'LandingPage')`

Please let me know if you have any questions, and thank you for your help!


Ravenstine:
You can automate adding the permission by creating a migration; it can have `Permission.create(resource: 'LandingPage')` under the `#up` method and `Permission.where(resource: 'LandingPage').destroy_all` under `#down`.

db/schema.rb
```diff
@@ -498,6 +498,30 @@
   add_index "flatpages_flatpage", ["path"], name: "django_flatpage_url", using: :btree
   add_index "flatpages_flatpage", ["updated_at"], name: "index_flatpages_flatpage_on_updated_at", using: :btree
 
+  create_table "landing_page_events", force: :cascade do |t|
```
Ravenstine: Since the concept of landing pages is fairly general, it might be worth considering making this table polymorphic so it can support other models.  I have a feeling someone in the future is going to ask to add things besides events.

arellaTV:
@Ravenstine 

Notes:
- `landing_page_events` is changed to `landing_page_contents` and is polymorphic. 
- There is validation for records without a slug
- Adding the LandingPage Permission is now a migration that uses the #up and #down methods like you highlighted earlier

Things I'm saving for post-mvp:
- Currently, the outpost aggregator only shows `events` as that will probably be its closest need. I imagine it would be nice if I can figure out how to make this a multi-select (possibly a multiselect input box that reinitializes the aggregator as the input changes).

arellaTV:
Thanks for taking a look @Ravenstine! Merging now

PR #983, opened 2017-10-20
Update to GTM

arellaTV:
Changelog:
- removes the previous Google Analytics script
- removes the function that checks for scroll depth (the scroll depth script is now in Google Tag Manager and loaded/invoked on DOMReady)
- populates the Data Layer with information that Patrick requested

I used Google Tag Manager's preview feature (which shows which tags are working) and the Google Analytics dashboard (to verify that events were being sent correctly) to test it out.


arellaTV:
Closing this for now until I make a more thorough audit of what we lose were we to disable vanilla GA.

PR #993, opened 2017-11-02
News Story and Program Segment cell caching

arellaTV:
- Implements caching mechanisms for cells that exist in both the news story and program segment pages

### DIFF
--- a/app/assets/javascripts/aplicayshun.js
+++ b/app/assets/javascripts/aplicayshun.js
@@ -11,6 +11,7 @@
 //= require cookie
 //= require visual_campaign
 //= require svgxuse/svgxuse.min
+//= require spin
 
 //= require embeditor
 //= require embeditor/templates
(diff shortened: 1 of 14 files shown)


arellaTV:
Seems there's a side effect with the ad caches (where they are displaced in the layout). Same goes for the social tools cell. Currently investigating

app/cells/category_cluster_cell.rb
```diff
@@ -12,7 +16,7 @@ def show &block
   end
 
   def content
-    @content ||= (model.try(:content) || []).map{ |a| a.get_article }.first(3)
+    @content ||= (model.try(:content) || []).try(:first, 3).try(:map, &:to_article)
```
Ravenstine: Why would we switch to using #to_article?  #get_article is essentially a cached version of #to_article, first checking for the indexed article in Eleasticsearch before trying to create it on the spot.

PR #1002, opened 2017-11-07
Update outpost previews

arellaTV:
Outpost previews are copies of normal views (with a few cells removed for performance purposes).

Example: https://scprv4-staging.scprdev.org/outpost/news_stories/69215/edit

### DIFF
--- a/app/controllers/outpost/events_controller.rb
+++ b/app/controllers/outpost/events_controller.rb
@@ -37,6 +37,9 @@ class Outpost::EventsController < Outpost::ResourceController
 
   def preview
     @event = Outpost.obj_by_key(params[:obj_key]) || Event.new
+    @more_events = Event.kpcc_in_person.upcoming.where("id != ?", @event.id).limit(2)
+    @past_events     = Event.kpcc_in_person.past.limit(5)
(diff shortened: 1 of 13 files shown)


app/controllers/outpost/events_controller.rb
```diff
@@ -37,14 +37,17 @@ class Outpost::EventsController < Outpost::ResourceController
...
+    @landing_page = LandingPage.find_by(title: 'KPCC In Person')
 
     with_rollback @event do
       @event.assign_attributes(params[:event])
 
       if @event.unconditionally_valid?
         @title = @event.to_title
         render "events/_event",
-          :layout => "outpost/preview/application",
+          :layout => "outpost/preview/new/application",
```
Ravenstine: Let's make it a point to not keep old code around while using directories prefixed with "new". This is something we need to avoid doing unless we're going to make a genuine cleanup effort(which almost never happens at KPCC).  Our tendency is to sweep old code under the rug, so it's best we limit doing this in the first place.

arellaTV: 👍

PR #1004, opened 2017-11-07
Shows archive description when an event has passed

arellaTV:
The changes are in `event_cell`. Logic is that it checks if the event is in the past, and if it is, to display the archive description instead of the body.

Unrelated sections of this PR:
- Removed vestigial template `outpost/preview/new/application`


arellaTV:
Needs to solve a test first. Will update here when that's ready

nakedsushi:
After tests are passing :shipit:

PR #1006, opened 2017-11-07
Enables listing and linking as it was in pre-AU

arellaTV:
Sorry for all the file changes! Had to add the method to every cell that needed it.

The new method first checks if there is an `original_object` from the input object, and uses that first. Otherwise, it uses the input object directly. This handles both cases where the data might not be in the right shape (might not be an article). Inside the code block, it checks whether that bio `is_public`, and if it is, creates a link, or else it just displays the `display_name`

Example: https://scprv4-staging.scprdev.org/news/2017/11/03/77364/amber-alert-chevy-truck-fort-tejon-gomes/

### DIFF
--- a/app/cells/archive_tab/blogs.erb
+++ b/app/cells/archive_tab/blogs.erb
@@ -8,7 +8,7 @@
     <h3 class="o-archive-tab__item-heading b-heading--h4 b-heading--serif"><%= link_to model.short_headline, model.public_path, class: 'b-link' %></h3>
     <span class="o-archive-tab__item-category"><%= link_to model.blog.name, blog_path(model.blog.slug), class: 'b-link' %></span>
     <span class="pipe">|</span>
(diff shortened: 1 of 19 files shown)


arellaTV:
If prod behaves like staging, it seems like you'd need to update the record in outpost to get it to use the new bylines.
Example of cached version with record not yet saved to retrigger the cache: https://scprv4-staging.scprdev.org/news/2017/11/03/77293/7-things-to-do-this-weekend-in-socal/

arellaTV:
The links above may not work as I'm currently using staging on another branch. Will update this post when it's back on.

arellaTV:
okay, the link above should work again

arellaTV:
Hoping to circle back to this soon and just create a helper function that all these different cells can use!

arellaTV:
Hoping to circle back to this soon and just create a helper function that all these different cells can use!

PR #1010, opened 2017-11-08
Fix listen live reporting

arellaTV:
- A `<title>` tag from an svg was changing what was being reported to chartbeat
- Will eventually move these svgs out of here and into `scprsprite.svg`


Ravenstine:
This looks good to me.  What a strange bug!  Also, an alternative to adding the SVGs to scprsprite would be to just remove their title tags... they're not necessary at all.

PR #1011, opened 2017-11-08
Fixes cache key to be more specific for recent content

arellaTV:
Fixes this bug: 

image

- Includes the class inside the cache_key to make sure these cells don't display without their necessary classnames


arellaTV:
@Ravenstine Sorry for the lack of clarity. When I say class, I mean CSS classname.

PR #1012, opened 2017-11-08
Fix short list and team photos

arellaTV:
Meant to fix this issue (when visiting http://www.scpr.org/short-list/2017/11/03/1837/weekend-reads/):

screen shot 2017-11-08 at 12 17 09 am

- Now the latest and short_list methods of the edition controller point to the same template, `editions/short_list` like it used to
- An adjacent fix for team photos are in the meet the team cell


nakedsushi:
LGTM :shipit:

PR #1014, opened 2017-11-08
Fixes breaking news access errors

arellaTV:
screen shot 2017-11-08 at 12 53 58 pm

The properties in the cell's template weren't being included. By adding `model.try()` to all of the necessary properties, there shouldn't be anymore access errors. Another way to do this would be to put includes at the top of the cell's controller. For now though, I think this way should be sufficient.

This is on me for not ever testing this feature. Since there isn't a mock for this (as far as I know), I'm importing the old stylesheet.

### DIFF
--- a/app/cells/breaking_news_alert/show.erb
+++ b/app/cells/breaking_news_alert/show.erb
@@ -1,9 +1,9 @@
 <div id="breaking-news">
-    <div class="live-update <%= alert_type %>">
+    <div class="live-update <%= model.try(:alert_type) %>">
       <div class="l-container inner-container">
           <div class="l-row">
             <div class="l-col--sm-12">
-              <strong><%= break_type %></strong><%= link_to headline, alert_url %>
+              <strong><%= model.try(:break_type) %></strong><%= link_to model.try(:headline), model.try(:alert_url) %>
               <a href="http://info.americanpublicmediagroup.org/LP=238" class="newsletter-sign-up">Sign up for alerts</a>
             </div>
           </div>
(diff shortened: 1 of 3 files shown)


app/cells/breaking_news_alert/show.erb
```diff
@@ -1,9 +1,9 @@
 <div id="breaking-news">
-    <div class="live-update <%= alert_type %>">
+    <div class="live-update <%= model.try(:alert_type) %>">
```
Ravenstine: Not that this won't work, but why would the breaking news alert cell ever receive an object that doesn't respond to those methods?

arellaTV: @Ravenstine It's more so that the previous way defaults to looking for an instance method called `alert_type` instead of first looking at the model.

Perhaps the better way to do it is like this?
property :alert_type
property :break_type
property :headline
property :alert_url

Ravenstine: @arellaTV I'm saying that theoretically, your breaking news alert cell should only ever receive a BreakingNewsAlert object, which means that calling #try for each property shouldn't be necessary.  It doesn't particularly hurt, but it shouldn't be needed since you're checking for `nil` before instantiating the cell.  The way you suggested is one way for that to not happen, but also rendering the cell without passing an actual BreakingNewsAlert object makes no sense and wouldn't even happen with the way it's written.

arellaTV: Yeah that's true! It'll always be in a position where it'll get a valid alert object. I think I'll lean to this just for now and come back to it once there's time for some code clean up. Thanks Ben! :)

PR #1019, opened 2017-11-09
Fixes contributing bylines by checking if theres an original_object

arellaTV:
Sometimes the current model doesn't have a `joined_bylines` property. In that case, check if there's an `original_object` property and use the `joined_bylines` from there instead.

### DIFF
--- a/app/cells/article_cell.rb
+++ b/app/cells/article_cell.rb
@@ -24,7 +24,8 @@ class ArticleCell < Cell::ViewModel
   end
 
   def biographies links=true
-    elements = model.try(:joined_bylines) do |bylines|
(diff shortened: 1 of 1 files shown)


app/cells/article_cell.rb
```diff
@@ -24,7 +24,8 @@ def audio_nav
   end
 
   def biographies links=true
-    elements = model.try(:joined_bylines) do |bylines|
+    original_object = model.try(:original_object) || model
```
nakedsushi: Is the order intentional here?  I think this logic says "Let the original object be the model's original object. If not, then just be the model."

PR #1022, opened 2017-11-09
Update editions.erb

tinykb:
Changing copy on Short List signup box: "Catch up with" to "Sign up to get"


arellaTV:
LGTM! :shipit:

PR #1038, opened 2017-11-16
Add meta_tag template to article cell

arellaTV:
Rather than use the method `render_content` to render the meta tags, I figured it would make sense to include that template as a method to the article cell. We can then use cell caching to cache the tags. 

Prior to this, `render_content` would check if there was a cache for a particular story, and if not, call `to_article` again on the input model to get it ready for the opengraph template. Since the model is usually in the correct shape by the time it reaches the article_cell, we wouldn't need to re-call `to_article` in the new implementation.

Also, rather than using `h()` to escape html tags, I elected to use `.to_json` like in the comment cell. Doing so meant removing the quotes that wrapped around the variables (ie `content=<%= strip_tags(model.teaser).to_json %>` instead of `content="<%= strip_tags(model.teaser).to_json %>"`)

### DIFF
--- a/app/cells/article/meta_tags.erb
+++ b/app/cells/article/meta_tags.erb
@@ -1,0 +1,29 @@
+<meta property="og:description" content=<%= strip_tags(model.teaser).to_json %>>
+<meta property="og:title" content=<%= model.short_title.to_json %>>
+<meta property="og:type" content="<%= @OG_TYPE || 'article' %>">
+<meta property="og:url" content="<%= model.public_url %>"/>
+<meta name="twitter:card" value="summary_large_image">
+<meta name="twitter:site" value="@kpcc">
+<meta name="twitter:url" value="<%= model.public_url %>">
+<meta name="twitter:title" value=<%= model.short_title.to_json %>>
+<meta name="twitter:description" value=<%= strip_tags(model.teaser).to_json %>>
+<meta property="article:published_time" content="<%= model.public_datetime %>" />
+<meta property="article:modified_time" content="<%= model.updated_at %>" />
+
+<% if model.category %>
+  <meta property="article:section" content=<% model.category.title.to_json %> />
+<% end %>
+
+<% if model.assets.any? %>
+  <% model.assets.map(&:full).each do |image| %>
+    <meta property="og:image" content="<%=image.url%>" />
+    <meta property="og:image:type" content="image/jpeg" />
+    <meta property="og:image:width" content="<%=image.width%>" />
+    <meta property="og:image:height" content="<%=image.height%>" />
+  <% end %> <%# assets %>
+  <link rel="image_src" href="<%= model.asset.full.url %>" />
+  <meta name="twitter:image" content="<%= https_to_http(model.asset.full.url) || "https://scpr.org/assets/kpcc-twitter-logo.png" %>">
+<% else %>
+  <meta property="og:image" content="<%=image_url('meta/facebook.jpg')%>" />
+  <meta name="twitter:image" content="<%=image_url('meta/twitter.jpg')%>">
+<% end %>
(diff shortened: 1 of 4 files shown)


app/cells/article/meta_tags.erb
```diff
@@ -0,0 +1,29 @@
...
+<meta name="twitter:description" value=<%= strip_tags(model.teaser).to_json %>>
+<meta property="article:published_time" content="<%= model.public_datetime %>" />
+<meta property="article:modified_time" content="<%= model.updated_at %>" />
+
+<% if model.category %>
+  <meta property="article:section" content=<% model.category.title.to_json %> />
+<% end %>
+
+<% if model.assets.any? %>
+  <% model.assets.map(&:full).each do |image| %>
```
nakedsushi: I have a question about this part. If the model has more than 1 image, then wouldn't the last image of this array end up overwriting the meta tags here? Or can we have more than 1 `og:image*` type tag per page?

If that's the case (only 1 opengraph image per page), should we just get the first image and generate the meta image tags just for that? It would save a few iterations on articles that have a ton of assets.

app/cells/article_cell.rb
```diff
@@ -23,6 +27,14 @@ def audio_nav
     render
   end
 
+  def meta_tags
+    render
+  end
+
+  def https_to_http url
+    url.try(:gsub, "https:", "http:")
```
Ravenstine: :notsure: why we would want to do this. 🤔

app/cells/article_cell.rb
```diff
@@ -23,6 +27,14 @@ def audio_nav
     render
   end
 
+  def meta_tags
+    render
+  end
+
+  def https_to_http url
```
nakedsushi: Why do we have this?

app/cells/article/meta_tags.erb
arellaTV: @nakedsushi Yes, we can have more than one og:image tag per page. In facebook, when you share, it shows up as a slideshow, and you can pick which one you want to use.

app/cells/article_cell.rb
arellaTV: @Ravenstine I copied this from the previous implementation. I think it was to fix a bug where twitter wouldn't show images with https as the protocol (but weirdly http worked)

app/cells/article_cell.rb
arellaTV: @nakedsushi I copied this from the previous implementation. I think it was to address a bug where twitter cards wouldnt show the image if the protocol was https (but weirdly would do it for http)

app/cells/article_cell.rb
Ravenstine: Wanna know what @nakedsushi thinks, but we should try out our HTTPS asset URLs with Twitter again to see if this is still an issue or if we can remove this code.  It's possible that ops took care of something around that but I don't recall.

app/cells/article/meta_tags.erb
arellaTV: This was also a copy from the previous implementation (probably to give the newsroom/web team more options for sharing). If it seems like the cons outweigh the pros (articles with a large number of assets, like https://www.scpr.org/news/2017/11/02/77121/pictures-of-dia-de-los-muertos-around-the-world/), then I would be open to just serving the first asset as the share image.

app/cells/article_cell.rb
nakedsushi: Yeah I *think* this is fixed now that our cert is valid, so let's test out not calling this method.

app/cells/article/meta_tags.erb
nakedsushi: Ah got it. Thanks for the explanation.

nakedsushi:
:shipit: and maybe create an issue to get rid of the https_to_http method

PR #1045, opened 2017-11-17
Portrait hero

arellaTV:
This change will automatically display the hero asset as an inline asset if it is a portrait sized asset.


arellaTV:
Closing this for now to give some room to think about a better solution

PR #1052, opened 2017-11-22
Caches cells in vertical pages

arellaTV:
There is a caching mechanism for each cell of the vertical template. Some cells (like the ipad appeal) don't necessary rely on the `@vertical` instance variable. In that case, I just cache it for 12 hours. Will update here with performance numbers.


arellaTV:
Closing this for now until a better solution is reached.

PR #1064, opened 2017-12-05
Aesthetic Unity: Speed revision 01

arellaTV:
These changes are in response to memory bloat issues that have been occurring recently. After some investigation, I recognized that active record transactions were really long. For verticals, it would average +20 seconds. The newest changes in this branch bring that down to around 80ms without caching.

There's another branch that has further optimizations site-wide that improves performance of news stories and segments down to 10ms range per request (https://github.com/SCPR/SCPRv4/tree/reduce-activerecord-usage), but it diverged so much (66 file changes) that I thought I would start over and create a branch with the most important changes first, then slowly add the rest of the optimizations later.

The general idea is that `to_article` is a bit too expensive, and being called too often in conjunction with many maps. The general process is to reduce the use of it wherever possible.

### Cells
A few cells have been altered to become more flexible (as in they allow for both Article models and native models such as NewsStories, Segments, etc.). Whenever there was a call for `short_title`, there's also a call for a `short_headline` (the native equivalent).

### Concerns
The change in this file offered the biggest speedup. If I'm understanding correctly, removing `to_article` basically stopped it from calling `related_content` in outgoing and incoming references as well.

### Models
These changes mainly affect vertical pages and give a substantial boost as `to_article` doesn't have to be run for every item in the collection anymore.

### Views
Because `related_content` can now be different types of models, it'll need to be conformed back to an Article model in outpost for it to work with the outpost_aggregator.

### Specs
The specs were updated to reflect the above changes.

### Compare
Branch in staging: https://scprv4-staging.scprdev.org/health
Production: https://scpr.org/health

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -47,13 +47,14 @@ class Tag < ActiveRecord::Base
       outgoing_references
         .order("position ASC")
         .where(related_omissions)
+        .limit(3).map(&:related)
     elsif taggings.count > 2
       taggings
         .order("created_at DESC")
         .where(tagging_omissions)
         .limit(10)
         .map(&:taggable)
-        .select{|a| a.respond_to?(:to_article)}
+        .select{|a| a.try(:public_datetime)}
         .first(3)
     else
       []
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -47,7 +47,6 @@ class Tag < ActiveRecord::Base
       outgoing_references
         .order("position ASC")
         .where(related_omissions)
-        .limit(3).map(&:related).map(&:to_article)
     elsif taggings.count > 2
       taggings
         .order("created_at DESC")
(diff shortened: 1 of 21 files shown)


app/models/tag.rb
```diff
@@ -47,16 +47,13 @@ def featured_content omit=[]
       outgoing_references
         .order("position ASC")
         .where(related_omissions)
```
nakedsushi: You may want to add back the `.limit(3)` or else this would grab all of the records that satisfy the `where` clause.

app/models/tag.rb
```diff
@@ -47,16 +47,13 @@ def featured_content omit=[]
...
-        .limit(3).map(&:related).map(&:to_article)
     elsif taggings.count > 2
       taggings
         .order("created_at DESC")
         .where(tagging_omissions)
         .limit(10)
         .map(&:taggable)
         .select{|a| a.respond_to?(:to_article)}
```
nakedsushi: Why was the `.select{|a| a.try(:public_datetime)}` removed here?

app/models/tag.rb
arellaTV: @nakedsushi Thank you! I actually just caught that after testing BetterHomepage saving. should be back now.

arellaTV: Also didn't mean to take out related_omissions. That should be on there as well

PR #1065, opened 2017-12-06
Aesthetic Unity: Speed Revision 02

arellaTV:
#These changes are in response to memory/cpu issues that have been occurring recently. `ProgramController#segment` is currently the slowest call according to NewRelic (averaging 607ms). The changes should lower transaction time (dropped to slightly under 100ms in staging), and cut ActiveRecord calls almost in half (from ~130 to ~77 according to Oink).

The basic idea is that the segment template should now only be fed the original object, which was actually how it was pre-unification. Before this PR, I had been using `@article = @segment.get_article` in the controller, and before that, `@article = @segment.to_article`, because I wanted to the shape of the data to fit the ArticleCell. However, I hadn't realized that there was a performance cost to `to_article`, likely related to the `related_content` association concern. This, coupled with `to_article` being called for every item of different collections, made for a lot of unnecessary operations.

### Cells

A few cells have been altered to become more flexible (as in they allow for both Article models and native models such as NewsStories, Segments, etc.). Whenever there was a call for short_title, there's also a call for a short_headline (the native equivalent).

### Controllers
The change in this file lowered transaction time for the controller itself, making it much lighter since it doesn't have to perform a `to_article` to get the segment ready for the segment template.

### Views
Because we're feeding the template a `@segment` instead of a `@article`, we have to feed all the cells that need the article shape a segment shape. Also removed unused layout templates.

### Specs
Removed a spec here that no longer applies (because we aren't using the program layout anymore).


arellaTV:
Thank you! 🚢

PR #1067, opened 2017-12-08
Aesthetic Unity: Speed Revision 02

arellaTV:
These changes are in response to memory/cpu issues that have been occurring recently. `NewsController#story` is now the slowest call according to NewRelic (averaging 294ms). The changes should lower transaction time (dropped to slightly around 100ms in staging).

The basic premise here is the same as the segment PR from yesterday (remove to_article from the controller, feed the template the original object, eliminate to_article from maps in cells).


arellaTV:
Thanks for taking a look! Merged

PR #1071, opened 2017-12-11
Fix donation tracking

arellaTV:
Some GA attributes were missing for donation clicks in the masthead and footer. This PR adds them back with new specifications (per @patrickdougall).

The events used to look like this:
`Category: Menu`
`Action: Donate`
`Label: Donate`

Patrick wants the new events to look like the following:
`Category: Menu`
`Action: Donate`
`Label: {url path}`

Other changes:
- Removes event tracking instantiation from the homepage and outpost (since this is being done GTM side now)
- Removes caching on the masthead (so that the request.path isn't also cached, also performance isn't dramatically affected when removed).
- Fixes Listen Live event tracking by including the GTM snippet properly via `<%= render "shared/new/meta_header" %>`


arellaTV:
Thank you! Merging

PR #1078, opened 2017-12-28
Adds editing feature to pledge drive exit modals

arellaTV:
This change allows for the pledge drive modal to be edited via a flatpage.

### DIFF
--- a/app/views/programs/kpcc/segment.html.erb
+++ b/app/views/programs/kpcc/segment.html.erb
@@ -98,10 +98,10 @@ end %>
           </svg>
         </span>
         <div id='exit-modal__content'>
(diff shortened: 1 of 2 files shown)


app/views/programs/kpcc/segment.html.erb
```diff
@@ -87,7 +87,7 @@ end %>
 
 <%= cell :epilogue, @segment, class: "l-column--boundless", order: 1001 %>
 
-<% if @modal || @is_pledge_time %>
+<% if @modal || (@is_pledge_time && @modal) %>
```
nakedsushi: This seems redundant. For example, if @modal is falsey, then `@is_pledge_time && @modal` would evaluate to falsey anyway.  What was the reason for this change from `if @modal || @is_pledge_time` ?

Issue #585, opened 2016-07-16
Deleted programs need to address orphaned associations


Ravenstine:
AC - KpccPrograms, when destroyed, do not cause errors because of orphaned segments/episodes.

We recently had a failure in Outpost caused by a show segment that was associated to a recently deleted KpccProgram(Reveal).  This caused a nomethod error that took down the ShowSegments index in Outpost.

- [ ] we know that this ticket is necessary or not

PR #1085, opened 2018-01-04
Make programs destroy their child records upon destruction.

Ravenstine:
#585

Does what it says on the tin using Rails' `dependent: destroy` option.
Did the same thing for both KPCC programs and external programs.


Ravenstine:
@nakedsushi It's pretty rare that we do, but it happened once and it was an approved action when it happened.  Since we haven't had anyone delete a program on accident, I think it makes sense to implement this to prevent the site from breaking if such a thing happens, and we recover the data from backups.  I am open to "ghosting" the related records, but I doubted that it was worth the effort since there was only ever one program deleted in my entire time here at KPCC.


- uəᗺ

________________________________

PR #1092, opened 2018-01-10
Changes to support Repeat release

arellaTV:
These changes are in preparation for the Repeat marketing page. They mainly add functionality to Flatpages, improves sharing functionality, and adds any necessary/useful assets.

### Assets
Two libraries are included here for scrolling/lazy loading functionality. The current version of the Repeat marketing page hasn't yet leveraged lazy loading, but I think it'll be an important option to have in future projects that are asset heavy. Code-wise, I made some changes to the way events are being bound to social icons. Before this, the `utilities.js.coffee` would select only one social icon for each of facebook, twitter, and email. In the changes below, it now applies an `each` so that multiple facebook/twitter/email links can be clicked on a single page.

### Controllers
This was a change to remove an old approach to serving flatpages (using a separate layout for full width flatpages). This seems unnecessary now because the `application` layout already provides the necessary scripts/stylesheets that we need, and it's basically identical to the layout that was used before `app_nosidebar`.

### Views
I added the ability to add an `extra_metahead` to flatpages. Code placed in the `extra_metahead` column is inserted into the header tag where other meta tags are. The related field in outpost is also added.

### DB
There were two migrations here. The first was to add an `extra_metahead` column to the flatpage model. The second was to change its type from `:string` to `:text`. I could have done this in one migration, but I didn't have enough foresight on the first one, and I didn't want to modify the migration after it was made.

### Public
Added the Apple Podcast icon and a separate wide-screen fallback image.

### Specs
The specs were updated to reflect the above change that flatpages are served within the `application` template instead of `no_sidebar` if the `full` option is selected.

### Vendor
Imports placed for scrolling and lazy loading jquery libraries.

Branch in staging: https://scprv4-staging.scprdev.org/repeat
Production: Upcoming


app/assets/javascripts/utilities.js.coffee
```diff
@@ -177,24 +177,27 @@ class scpr.SocialTools
...
-        $(@options.twitfinder).on "click", (evt) =>
-            if url = $(@options.twitfinder).attr("data-url")
-                headline = $(@options.twitfinder).attr("data-text")
-                twurl = "https://twitter.com/intent/tweet?" +
-                        "url=#{url}&text=#{headline}&via=kpcc"
-                window.open twurl, 'pop_up',
-                    'height=350,width=556,' +
-                    'resizable,left=10,top=10,scrollbars=no,toolbar=no'
-            false
+        $(@options.twitfinder).each (index, el) =>
```
nakedsushi: Just checking, but this should work even if there's only one element with a class that matches `@options.twitfinder`, right?

app/assets/javascripts/utilities.js.coffee
```diff
@@ -177,24 +177,27 @@ class scpr.SocialTools
...
-                emurl = "/content/share?obj_key=#{key}"
-                window.open emurl, 'pop_up',
-                    'height=830,width=545,' +
-                    'resizable,left=10,top=10,scrollbars=no,toolbar=no'
-            false
+        $(@options.emailfinder).each (index, el) =>
+          $(el).on "click", (evt) =>
+              if key = $(el).attr("data-key")
+                  emurl = "/content/share?obj_key=#{key}"
+                  debugger
```
nakedsushi: Should this be here?

app/views/shared/new/_meta_header.html.erb
```diff
@@ -51,3 +51,45 @@
 <%= amp_head_link %>
 
 <%= favicon_link_tag '/favicon.ico' %>
+
+<script>
+  window.fbAsyncInit = function() {
+	FB.init({
+	  appId            : '162204204509247',
```
nakedsushi: Double checking: this is our production app_id?

arellaTV: @nakedsushi Regarding the two migrations, they've only been done on staging so far (not yet on prod).

app/assets/javascripts/utilities.js.coffee
arellaTV: Yes, this way also works if there's only one element that matches. in that case, it would be a jquery object with an array of length 1.

app/assets/javascripts/utilities.js.coffee
arellaTV: AH! No, it should not! Will take this out shortly. Thanks for catching that!

app/views/shared/new/_meta_header.html.erb
arellaTV: @nakedsushi This is a separate app_id from the one we've been using. I reached out to Bryan Ricker to see if he was the one who set up the app, but he was not. He said it might be Sean Dillingham or Eric Zassenhaus. Should I proceed with asking them if they have access to the app?

db/migrate/20180116201203_add_extra_metahead_to_flatpages.rb
```diff
@@ -0,0 +1,9 @@
+class AddExtraMetaheadToFlatpages < ActiveRecord::Migration
+  def up
+    add_column :flatpages_flatpage, :extra_metahead, :text, :limit => 4294967295
```
nakedsushi: Is there a reason you chose that limit? I think the default, which is something like 64512 bytes would be sufficient for the extra metaheads? If it's not, we should rethink what's being stored in this column (like, should it be in a template instead?).  Further reading: https://stackoverflow.com/questions/13932750/tinytext-text-mediumtext-and-longtext-maximum-storage-sizes/35785869#35785869

arellaTV: Hi @nakedsushi! Yes, I chose that limit to match the limits of the `extra_head` and `extra_tail`. But I agree, the meta head doesn't need that much, so I think I can stick to the default for text for now. Will update this in a sec!

arellaTV:
Given the go ahead to :shipit:

PR #1104, opened 2018-01-19
Sets the global expiration for the cache to one week

arellaTV:
From the `dalli_store` docs, you can set a global expiration: https://github.com/petergoldstein/dalli#configuration
> **expires_in:** Global default for key TTL. Default is 0, which means no expiry.

This will take care of pages with mixed content errors, e.g. https://www.scpr.org/education, because those were stored in the cache before the 1 minute expiration write logic was put into effect.

screen shot 2018-01-17 at 1 54 39 pm

This is in conjunction with another PR that adds expirations to cache reads: https://github.com/SCPR/asset_host_client/pull/2

Note: Pages with ads will continue to have mixed content errors because ads are served with the `http` protocol for some reason. Will create another ticket to address this in the future.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -3,7 +3,7 @@ Scprv4::Application.configure do
   config.eager_load     = true
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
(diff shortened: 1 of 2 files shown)


nakedsushi:
:shipit:

PR #1105, opened 2018-01-24
Fix audio params

arellaTV:
So sorry for the huge PR!

The basic gist is that anywhere we input an audio file's url, we replace that with the method `url_with_params({audio url goes here}, context: {'news' || program slug, blog slug}, via: 'website')`. I also made a conscious decision to make `context: outpost` for outpost previews. If it makes more sense to leave params off altogether for these, I can change them back.

I tested each of these changes individually with the network tab in chrome dev tools. This branch is up on staging, e.g.
https://scprv4-staging.scprdev.org/programs/the-frame/2018/01/20/61296/a-documentary-goes-behind-the-scenes-with-rbg-ruth/

There are also a couple of adjacent bug fixes, specifically:
- Adding a title to the header that was missing for episode pages
- Fixing an overflow issue in the featured story of a program page

### DIFF
--- a/app/cells/article_audio/horizontal.erb
+++ b/app/cells/article_audio/horizontal.erb
@@ -1,6 +1,6 @@
 <div class="c-audio--horiz">
   <div class="c-audio-widget">
-    <a href="<%= audio_file.url %>" class="c-play__button c-audio__button" title="<%= model.try(:title) %>" data-duration="<%= audio_file.duration %>"></a>
+    <a href="<%= audio_url audio_file.url %>" class="c-play__button c-audio__button" title="<%= model.try(:title) || model.try(:headline) %>" data-duration="<%= audio_file.duration %>"></a>
   </div>
   <div class="c-audio__text">
     <p class="c-audio__text-listen">Listen to <%= @options[:type] %></p>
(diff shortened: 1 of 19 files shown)


app/cells/episode_cell.rb
```diff
@@ -1,4 +1,5 @@
 class EpisodeCell < Cell::ViewModel
+  include ApplicationHelper
```
nakedsushi: Why is ApplicationHelper included here?

app/cells/article_audio/horizontal.erb
```diff
@@ -1,6 +1,6 @@
 <div class="c-audio--horiz">
   <div class="c-audio-widget">
-    <a href="<%= audio_file.url %>" class="c-play__button c-audio__button" title="<%= model.try(:title) %>" data-duration="<%= audio_file.duration %>"></a>
+    <a href="<%= audio_url audio_file.url %>" class="c-play__button c-audio__button" title="<%= model.try(:title) || model.try(:headline) %>" data-duration="<%= audio_file.duration %>"></a>
```
nakedsushi: Just to confirm, in cases like this, the resulting url will have no context because it's not being passed in to `audio_url()` ?

app/cells/article_audio_cell.rb
```diff
@@ -41,4 +42,8 @@ def audio_size
     end
   end
 
+  def audio_url url
+    url_with_params(url, context: @options[:audio_context], via: 'website')
```
nakedsushi: Just confirming, but if `@options[:audio_context]` is nil, this should still work, right?

app/cells/article_audio/horizontal.erb
arellaTV: @nakedsushi In this case, there will be a context as long as an `@option[:audio_context]` is passed into the cell. The method `audio_url` is an `article_audio_cell` method that only requires a url - inside that method, it's calling `url_with_params` with `@option[:audio_context]` as the context param.

app/cells/article_audio_cell.rb
arellaTV: @nakedsushi Yes, from my testing, it only adds params with non-nil values. So the resulting url would look something like:
`https://media.scpr.org/path/to/an_audio_file.mp3?via=website`

app/cells/episode_cell.rb
arellaTV: @nakedsushi I included ApplicationHelper here to add url_with_params so I didn't have to redefine it again at the cell level. In hindsight, it's probably not performant to include the whole set

nakedsushi: Ah, if you just want to include it for that method, does this work?
`include ApplicationHelper.url_with_params`

nakedsushi:
Other than that change w/ including ApplicationHelper, :shipit: but note my push/deploy notes.

arellaTV:
From what I can find, accessing a helper method this way:
`ApplicationHelper.url_with_params(args)` is not the same as `include ApplicationHelper`, which exposes every helper method as an action to the cell's controller. https://stackoverflow.com/a/5130344

Looks like this is the better approach. Thank you Louise! I'll be testing it out tonight

PR #1107, opened 2018-01-29
Adds GTM Data Layer logic to Audio Player Events

arellaTV:
Changes proposed:
- `addToDataLayer` method is used to push to the dataLayer so that GTM scripts can access it
- adds `play`, `pause`, and `change` events to the dataLayer
- `change` event occurs when the current audio source being loaded is different from its immediate predecessor
- added comments/explanation for changes

To test:
- Go to GTM and turn on "Preview" mode
- Visit this page: https://scprv4-staging.scprdev.org/programs/take-two/2017/12/01/60481/cruising-through-the-2017-la-auto-show/
- Choose the dataLayer tab
- Click the play button
- A dataLayer object with `eventAction: play` should show up
- Minimize the GTM dev tool (because it's blocking the audio bar)
- Press pause, and then click the second audio element (or click the second audio element and then click pause)
- Maximize the GTM dev tool
- Two more dataLayer objects with `eventAction: pause` and `eventAction: change` should show up

### DIFF
--- a/app/assets/javascripts/README.md
+++ b/app/assets/javascripts/README.md
@@ -1,0 +1,58 @@
+# JavaScript Assets
+
+The JavaScript files in this directory serve as a way to enable dynamic interaction across the site. `aplicayshun.js` currently holds all the scripts required globally. You can find usage details for some major scripts below.
+
+---
+
+## `audio.js.coffee`
+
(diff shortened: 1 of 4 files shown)


Ravenstine:
The code looks good to me.  Unless there's a rush to deploy this, I think it'd be helpful to copy your testing instructions down to a markdown file in the JavaScripts directory.  Even if it's just those bulletpoints, I'm pretty sure Louise and myself are less GTM-savvy so it might save us some time down the road.

PR #1111, opened 2018-02-03
Simplifies aspect logic and removes dark grey bg

arellaTV:
Changelog:
- Removes unnecessary css class declarations
- Simplifies aspect logic
- Removes grey background from portrait images


arellaTV:
After looking at this again, I realized that the logic is incomplete. Change incoming!

PR #1122, opened 2018-02-13
Fixes podcast link bug for external program pages

arellaTV:
The previous implementation of presenting podcast links in external pages was partially hardcoded. The changes here now get the proper links, `podcast` and `rss`, and shows them in the cell if they are available. If neither are available, the cell will not render. Assumptions are tested in the cell spec linked below.

**Documentation:** https://github.com/SCPR/SCPRv4/blob/bbe1ee345e830fe4564440aebbe67110c3f88129/app/cells/appeal/README.md

**Tests:** https://github.com/SCPR/SCPRv4/blob/bbe1ee345e830fe4564440aebbe67110c3f88129/spec/cells/appeal_cell.rb

**Staging demo:**
(XML) https://scprv4-staging.scprdev.org/programs/1a
(Neither) https://scprv4-staging.scprdev.org/programs/reveal/

### DIFF
--- a/app/cells/appeal/podcast.erb
+++ b/app/cells/appeal/podcast.erb
@@ -2,11 +2,11 @@
   <h4 class="o-appeal__heading b-heading b-heading--h4 b-heading--serif b-heading--bold b-heading--italic">Subscribe to<br /><%= model.title %>'s podcast</h4>
   <p class="o-appeal__caption">Receive every new episode, automatically, as soon as it's released.  Listen on your own time, your own way.</p>
   <% present model, ProgramPresenter do |p| %>
-    <% if p.send("podcast_link").present? %>
-      <a class="o-appeal__submit c-btn c-btn--outline-secondary c-btn--small" href="<%= p.send("podcast_link") %>">Subscribe via iTunes</a>
+    <% if !p.podcast_link.blank? %>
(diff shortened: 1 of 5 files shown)


app/cells/appeal/podcast.erb
```diff
@@ -5,7 +5,9 @@
     <% if p.send("podcast_link").present? %>
       <a class="o-appeal__submit c-btn c-btn--outline-secondary c-btn--small" href="<%= p.send("podcast_link") %>">Subscribe via iTunes</a>
     <% end %>
+    <% if p.send("rss_link").present? %>
```
nakedsushi: Question, why do `p.send("rss_link")` instead of `p.rss_link` ?

arellaTV: @nakedsushi 
I think I was mirroring how the presenter was used in the pre-unified version here: https://github.com/SCPR/SCPRv4/blob/684d1f8e10da6e55cad104f7f4d801a5929e3054/app/views/shared/new/masthead/_program_titling.html.erb#L9-L11

But looking at it now, I agree that it makes a lot more sense to just access the property directly, e.g. `p.rss_link`. On it now and will update when it's ready! Thank you Louise!

PR #1141, opened 2018-03-02
Removes the raw invocation on string properties and lets rails escape it

arellaTV:
There were some issues with the ingest process of our podcasts to Megaphone. I talked with a representative, and she said that most of the time the issue is that the output contains utf16 strings instead of utf8. 

### Removing raw() in the podcast XML builder
I looked at the XML builder and realized that we were calling [raw()](https://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/raw) on all of the string properties (i.e. `title`, `itunes:summary`, `itunes:author`, `description`, and `keywords`). By default, Rails automatically escapes tags, and we would want to do that for the output here as the tags won't be necessary on podcast clients.

After removing the raw invocations, shipping to staging, and supplying Megaphone with a staging podcast xml like https://scprv4-staging.scprdev.org/podcasts/airtalk, the ingest was able to complete successfully.

### UTF-8 concerns
I trust that `article.title`, `@podcast.author`, `article.teaser`, and `@podcast.keywords` are in proper utf-8 because thats what outpost outputs (I think..). Do you think I should go another step and apply a force encoding to these properties as well to ensure that they are truly utf-8 valid?


Ravenstine:
Do we know that we are currently UTF-16 characters?  I don't know how we would be doing that in the first place.  Our database doesn't even store UTF-16 IFAK.

arellaTV:
@Ravenstine I asked if they could provide an example of where it might be failing, but they didn't really answer that. I'd probably have to find it myself with some type of validator

arellaTV:
Closing this as it appears it was a false positive

PR #1147, opened 2018-03-08
Allow for a heterogenous list in ShowEpisode pages

arellaTV:
It used to be possible to include non-`ShowSegment` records in `ShowEpisode` pages. However, since AU, only `ShowSegments` have been showing up because it only pulls `ShowSegment` records in the cell (with `model.try(:to_episode).try(:segments)`. In this PR, we go back to using the `@content` instance variable that was supposed to be used to render this list, and make the cell more passive.

**Documentation:** https://github.com/SCPR/SCPRv4/blob/236b2c83bd071b742df37feb7932b21fe2c9a146/app/cells/episode/README.md

**Test:** https://github.com/SCPR/SCPRv4/blob/236b2c83bd071b742df37feb7932b21fe2c9a146/spec/cells/episode_cell.rb

**Staging demo:**
(Heterogeneous) https://scprv4-staging.scprdev.org/programs/take-two/2018/02/23/17893/
(Homogeneous) https://www.scpr.org/programs/take-two/2018/02/23/17893/

*Note: in the examples above, the staging link correctly shows the `NewsStory` record, "11 things to do this weekend in SoCal", whereas the production link doesn't

### DIFF
--- a/app/cells/episode/show.erb
+++ b/app/cells/episode/show.erb
@@ -12,8 +12,8 @@
     <div class="o-featured-episode__segment">
       <figure class="o-featured-episode__segment-figure"><img src="<%= asset_path segment %>" style="background-image:url(<%= asset_path segment %>);"></figure>
       <div class="o-featured-episode__segment-description c-audio-widget">
-        <h3 class="o-featured-episode__segment-title"><%= link_to segment.short_title, segment.public_path %></h3>
-        <div class="o-featured-episode__segment-teaser"><%= segment.teaser %></div>
+        <h3 class="o-featured-episode__segment-title"><%= link_to segment.try(:short_title) || segment.try(:short_headline), segment.try(:public_path) %></h3>
+        <div class="o-featured-episode__segment-teaser"><%= segment.try(:teaser) %></div>
         <% if audio_file(segment) %>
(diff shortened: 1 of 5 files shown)


app/cells/episode/show.erb
```diff
@@ -12,8 +12,8 @@
     <div class="o-featured-episode__segment">
       <figure class="o-featured-episode__segment-figure"><img src="<%= asset_path segment %>" style="background-image:url(<%= asset_path segment %>);"></figure>
       <div class="o-featured-episode__segment-description c-audio-widget">
-        <h3 class="o-featured-episode__segment-title"><%= link_to segment.short_title, segment.public_path %></h3>
-        <div class="o-featured-episode__segment-teaser"><%= segment.teaser %></div>
+        <h3 class="o-featured-episode__segment-title"><%= link_to segment.try(:short_title) || segment.try(:short_headline), segment.try(:public_path) %></h3>
```
nakedsushi: Was this to fix the problem where the "11 things to do this weekend" was breaking the BetterHompage updating?  IIRC the problem was that the news_story or whatever didn't have a `public_path()`? Or maybe the news_story never got passed in? 

Are we confident that whatever is in `segment` will have a `public_path()`?

arellaTV: Hi @nakedsushi! Yes, we should be confident that whatever `segment` is will have a `public_path`, as the collection itself goes through a `to_article` map (https://github.com/SCPR/SCPRv4/blob/56da24e8f2b399b6d75862c9be47dcd233ebc297/app/concerns/concern/controller/show_episodes.rb#L6), and articles have the `public_path` property. The `.try` was more of a precaution, but if it seems unnecessary, I can change it back  to just `segment.public_path`.

arellaTV:
Thanks @nakedsushi! Will merge, ship, and track for any issues

PR #1157, opened 2018-03-19
Changes around testing

Ravenstine:
Here's what I've done:

- Cleared out all migrations.  This might be controversial so I am fine rolling that commit back if need be, but I can't think of a single time where I needed to look back at old migrations.  At the very least, we should only keep around new migrations until it's clear we've committed to a change.  Otherwise, a new database should be built using `rake db:setup` or, more specifically, `rake db:schema:load`.
- Changed tests to rely on an existing Elasticsearch instance.  Originally, it was written to spin up its own Elasticsearch instance, presumably to avoid polluting the development instance.  This is not very convenient for a few reasons; it means managing both a native version and a Docker version of Elasticsearch and adding several seconds to boot time even for tests unrelated to Elasticsearch.  I am also not a fan of the application code spinning up services like that.  I changed the default namespace for indexes created by the tests so that, after the suite is completed, it will erase only the test indexes while leaving the development ones alone.
- Some methods for suppressing stdout were removed from spec_helper.rb because, starting in Ruby 2.3, $stdout and $stderr are frozen objects and can't be easily replaced.  Removing those methods all together doesn't appear to change the test output.  It's possible that it did at one point, but then again, our application shouldn't be printing anomalous strings to stdout anyway.  This issue was preventing me from running the tests locally.
- I removed restore-percona-backup as the default MySQL container in docker-compose.yml.  Depending on production backups is an anti-pattern to begin with, but our backups are so colossal that I can't dependably start a fresh instance of SCPR.org without either running out of disk space or freezing Docker.  It's ridiculous.  Instead, I replaced it with a regular MySQL image.  With a seeded superuser for Outpost, it's minimally sufficient for getting the application into a state suitable for development.  There should be more seed data, but we're at a point where an empty database is better than wasting hours on production backups.

The change to the MySQL configuration means that these are all the steps to run SCPR.org for the first time(assuming you have Ruby, Node, and Docker):

1. `git clone https://github.com/SCPR/SCPRv4.git`
2. `bundle install & npm install`
3. `docker-compose up mysql elasticsearch redis memcached`
4. `bundle exec rake db:setup`
5. `bundle exec rake db:seed`
6. `rails s`


spec/spec_helper.rb
```diff
@@ -164,21 +161,12 @@
...
-  @original_stderr = $stderr
-  @original_stdout = $stdout
-
-  # Redirect stderr and stdout
-  $stderr = File.new('/dev/null', 'w')
-  $stdout = File.new('/dev/null', 'w')
+
+def erase_test_indices
+  ContentBase.es_client.indices.segments["indices"].keys.each do |index|
+    if index.match(ES_ARTICLES_INDEX)
```
nakedsushi: Will this also erase dev articles index if dev is running the same ES container?

Ravenstine: No, because the default index name for testing is namespaced as "scprv4-test". It will only clear indexes with that prefix. Basically, it's replicating how databases are handled in testing. Although it's possible to have your indexes erased if your existing secrets.yml still has the old prefix. It wouldn't be the worst thing in the world because a developer shouldn't have so many local records that reindexing would be onerous.

arellaTV:
Looks good to me! @nakedsushi I don't rely on recent articles locally, so that shouldn't be a problem

PR #1174, opened 2018-04-19
Includes pym.js

nakedsushi:
We weren't including this in our individual article pages. We should, if we want the magic of responsive iframes.


arellaTV:
LGTM! :shipit:

nakedsushi:
Merged https://github.com/SCPR/SCPRv4/commit/a2128d6da19a2aa696043f1267625ec31f0ebd23

PR #1178, opened 2018-04-23
Automatically publish articles that are synced from NPR's API

arellaTV:
## Chain `.import` after `.sync`
The biggest change is in `app/importers/npr_article_importer.rb`.
- We call `.import` immediately after an article is `synced` successfully and "Watch Live" is not in the title (we use the regular expression `/watch\s+live/i`)
- In `.import`, rather than always making an API call, we check if an `npr_story` has already been passed through as an option, and if it has, use that instead (this is to reduce the amount of individual API calls we make and just rely on the batch call).
- `.import` tries to match the story's `primaryTopic` with one of our categories (uses the lookup table defined at the top called, `CATEGORY_MAP`.
- Also, instead of saving the story as a draft, it is immediately published

### Tests
https://github.com/SCPR/SCPRv4/compare/npr-auto-import?expand=1#diff-9bda9a793a03bdb329896f2d31778192
#### Added tests:
- It calls import after a successful sync
- It does not call import if the npr story is a live video, but still adds it as a RemoteArticle
- It adds a category if NPR's primaryTopic matches one of our categories

## Associate `Category` to `List`
The biggest changes are in the List model, List API controller, and List Outpost Controller.
- Lists now have a category association (uses the Category Association Concern).
- If a category is associated to a list, but there are no curated items, then the list api returns items from the associated category (de-duped from the latest 16 stories).
- If a category is associated to a list, but there are curated items, then the list api returns the curated items instead

### Migration
https://github.com/SCPR/SCPRv4/compare/npr-auto-import?expand=1#diff-38444cef60d9fb90036e4b8fdabba0ec

### Tests
https://github.com/SCPR/SCPRv4/compare/npr-auto-import?expand=1#diff-0f0f7f0a77d9b1b458b079b8891492fb
https://github.com/SCPR/SCPRv4/compare/npr-auto-import?expand=1#diff-aab21b6676e9786a2c7c8b1479b414e9
#### Added tests:
- It returns associated articles even if a category is associated
- It returns category items if there are no associated articles
- A list of news stories associated with a specific category is de-duped from the latest 16 news stories.


app/importers/npr_article_importer.rb
```diff
@@ -58,6 +71,11 @@ def sync
             added.push cached_article
             log "Saved NPR Story ##{npr_story.id} as " \
                 "RemoteArticle ##{cached_article.id}"
+            if npr_story.title =~ /watch\s+live/i
```
nakedsushi: For a slight efficiency gain, you can store `/watch\s+live/i` as a constant on the class (like you did with `CATEGORY_MAP`) and just do `if npr_story.title =~ WATCH_LIVE_REGEXP` (or something like that).  That way, the REGEXP is compiled only once, instead of each time `sync()` is called.

Since the regexp is pretty simple, it doesn't buy you too much, but it does save you from wading through the code if you have to change the REGEXP in the future since it'll be at the top w/ the other constants.

app/importers/npr_article_importer.rb
```diff
@@ -95,14 +113,18 @@ def import(remote_article, options={})
...
       # Build the NewsStory from the API response
       article = klass.new(
-        :status         => klass.status_id(:draft),
+        :status         => klass.status_id(:live),
         :headline       => npr_story.title,
         :teaser         => npr_story.teaser,
         :short_headline => npr_story.shortTitle.present? ? npr_story.shortTitle : npr_story.title,
-        :body           => text
+        :body           => text,
+        :category_id    => Category.find_by(title: CATEGORY_MAP[primary_topic_title]).try(:id)
```
nakedsushi: Will this fail gracefully if `primary_topic_title` is nil here?  Do we just build a new NewsStory with no category mapping?

app/models/list.rb
```diff
@@ -55,6 +57,14 @@ def build_item_association(item_hash, item)
     ListItem.new(attrs) if item.published?
   end
 
+  def deduped_category_items
+    self.category.content({
+      page: 1,
+      per_page: 14,
+      exclude: NewsStory.published.order('published_at DESC').limit(16)
```
nakedsushi: Is this so that we don't list any NewsStory's that are already in the Headlines?  If yes, then I wonder if we can make `NewsStory.published.order('published_at DESC').limit(16)` into a method that gets used here as well as where we expect to return Headlines. That way, if we end up changing the code that produces the Headlines newstories (for example, we want 20 instead of 16) we'll only need to change it in one place.

db/schema.rb
```diff
@@ -1048,6 +1049,83 @@
   add_index "tags", ["created_at"], name: "index_tags_on_created_at", using: :btree
   add_index "tags", ["slug"], name: "index_tags_on_slug", using: :btree
 
+  create_table "test_class_alerts", force: :cascade do |t|
```
nakedsushi: What are these schema changes for?

app/importers/npr_article_importer.rb
arellaTV: Ohhh, that makes sense! That sounds great, I'll add that into the next commit

app/importers/npr_article_importer.rb
arellaTV: @nakedsushi, yes, if primary_topic_title is nil, `Category.find_by` returns nil, which then just doesn't associate a category to the NewsStory.

db/schema.rb
arellaTV: @nakedsushi Hmmm, that was generated after I performed a migration on the scprv4_test database to get it to recognize the new column on lists, but it seems like we can do without the `test_class_alerts` portion (I'm not sure how that came up. I should investigate why it added that

app/models/list.rb
arellaTV: @nakedsushi Yes, that line is so that news stories don't show up both in the headlines list and in a list below it. That definitely makes sense, I'll go ahead and add that to my to-do for my next commit. Thank you!

Issue #560, opened 2016-06-29
Add Tagline to Politics Vertical footer


vsinghkpcc:
AC
- [x] The copy "No rant, no slant" appears in the footer of the politics landing page in place of "89.1 KUOR 90.3 KVLA"

Ravenstine:
@vsinghkpcc kpcc My guess is that the politics page is sharing a footer template with other pages.  To be clear, do we only want this tagline to show up on the politics page and no other pages, or all pages with that footer?

vsinghkpcc:
Let's limit to just the politics page, thanks!

PR #1181, opened 2018-04-25
Replace Embeditor w/ Portable Holes

Ravenstine:
#560

This replaces Embeditor with Portable Holes, which allows us to use the same embed-placeholder system between SCPR.org and Grand Central without the need for crazy precompilation steps, etc., and the Facebook embeds no longer need to use Embedly.

A lot of code around Embeditor has been removed because SPCR.org no longer has to run a specific JavaScript environment with CoffeeScript compilation just to be able to use Embeditor for AMP pages, etc.  It can simply run Portable Holes in Node.js.

The old application.js has been removed and aplicayshun.js has been renamed to application.js.

PouchDB has been removed as a dependency because it wasn't being used by anything, and it appears that our deployments are a little faster now as a result.  I also deleted unused code that once needed it.

Here is an example page on staging with both Facebook and YouTube embeds:
https://scprv4-staging.scprdev.org/news/2018/04/25/82494/giving-antibiotics-to-healthy-kids-in-poor-countri/


app/assets/javascripts/application.js
```diff
@@ -1,49 +1,33 @@
...
-//= require smart_time
-//= require event_tracking
-//= require visual_campaign
-//= require open_popup
-//= require utilities
-//= require collapse
-//= require cms_popup
-//= require reader_bookmarklet
-//= require page_mapping
+PortableHoles = require('portable-holes');
```
nakedsushi: Did you intend for PortableHoles to be a global?

package.json
```diff
@@ -35,8 +35,8 @@
     "najax": "^0.7.1",
     "normalize.css": "^4.1.1",
     "pacote": "^2.7.36",
+    "portable-holes": "^1.2.1",
     "pouchdb": "= 5.3.1",
```
nakedsushi: Can we get rid of pouchdb in here too?

app/cells/article/show.erb
```diff
@@ -44,12 +44,12 @@
     // Load In-Article Membership Appeal from AdHost
     scpr.VisualCampaign.enqueue('article-appeal', $('#article-membership-appeal'));
 
-    // Load Embeditor
-    embeditor = new Embeditor.Base({
+    // Load PortableHoles
+    holes = new PortableHoles({
       Embedly: {
         query: { key : '0cb3651dde4740db8fcb147850c6b555' }
```
arellaTV: Do we still need to provide an Embedly key in the off chance that we're using Embedly for embeds other than Facebook, Youtube, Twitter, and Instagram?

Ravenstine: Yeah, Embedly is still a fallback for embed providers we don't have adapters for.  Those include Vine, Vimeo, Google Maps, Scribd, Soundcloud, and Spotify.  I'll probably move even more of those away from Embedly over time, but Facebook and YouTube make up the bulk.  Even if we moved all of those, Embedly still can act as a "polyfill" of sorts for anything else.  I think we should keep using it, but not depend on it.

app/assets/javascripts/application.js
Ravenstine: Yes.  Embeditor was already a global.  Because of our future course corrections, I didn't want to change how it gets used.  The only difference is it's now an NPM module instead of a Ruby gem.

package.json
Ravenstine: Oh weird!  I removed pouchdb-upsert but not pouchdb.  I could swear I did.  Yup, I'll remove it tomorrow before we ship it.

PR #1182, opened 2018-04-25
Integrate Megaphone into Outpost

arellaTV:
**The main file to look at is: `app/models/show_episode.rb`.**
I added getters and setters for the following megaphone properties:
- `preCount`
- `postCount`
- `insertionPoints`
On save, if any of the above are changed, it executes a `put` request.

**Other features include:**
- Appends the `Rails.env` to the podcast rss output so that changes in local/staging don't affect production records.
- Only showing the "Podcast Ad Placement" widget on shows that are "onair"
- If the initial `get` is unsuccessful, it disables the widget and shows a notice that the podcast record is unavailable for editing.
- Form validation for the insertion point gives immediate feedback

**Testing**
- `spec/features/outpost/manage_show_episodes_spec.rb`
- `spec/models/content_base/show_episode_spec.rb`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,7 @@ gem 'audio_vision', '~> 1.0'
 gem 'slack-notifier'
 gem 'aws-sdk', '~> 2'
 gem 'one_signal'
+gem 'megaphone_client', github: "scpr/megaphone_client"
 
 ## Assets
 gem "eco", "~> 1.0"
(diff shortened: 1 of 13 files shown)


arellaTV:
Oops! Didn't realize there were tests failing. Please hold until I resolve.

arellaTV:
@Ravenstine and @nakedsushi, okay, looks like the tests are passing again. Please take a look whenever you are free.

PR #1200, opened 2018-05-15
Rescues errors that occur when pulling in from PMP

arellaTV:
Changes:

- rescues any error that occurs during a PMP API call

The error that keeps happening is a 504, which is usually a timeout error. I was able to perform the queries manually through rails console, and they are giving back valid responses, so for now we'll rescue by defaulting the `stories` variable to an empty array.


arellaTV:
Was given a go ahead to :shipit:

PR #1201, opened 2018-05-15
Adds a check for an on/off switch and uses that if it exists

arellaTV:
Changes:

- Looks for a `DataPoint` called "npr_auto_publish_on", and if it's set to 0, it turns off autopublishing, set to 1, it turns it back on.
- Moved the rails environment check to the `auto_publish` method because thats where other blocking conditions happen (such as `elligible_for_autopublish?`).

### DIFF
--- a/app/importers/npr_article_importer.rb
+++ b/app/importers/npr_article_importer.rb
@@ -69,7 +69,7 @@ module NprArticleImporter
       npr_auto_publish_on = DataPoint.find_by(data_key: "npr_auto_publish_on").try(:data_value).try(:to_i) || 1
 
       # Check that all conditions pass: auth-publish is on, the article is elligible, and the environment is either production or test
-      if !npr_auto_publish_on.zero? && elligible_for_autopublish?(npr_story) && (Rails.env == 'production' || Rails.env == 'test')
+      if !npr_auto_publish_on.zero? && elligible_for_autopublish?(npr_story) && (Rails.env.production? || Rails.env.test?)
         self.import remote_article, { npr_story: npr_story, manual: false }
       else
         log "Skipping auto-import of story: #{npr_story.id}"
--- a/app/importers/npr_article_importer.rb
+++ b/app/importers/npr_article_importer.rb
@@ -64,10 +64,15 @@ module NprArticleImporter
     end
 
     def auto_publish remote_article, npr_story
-      if elligible_for_autopublish?(npr_story)
+
+      # A boolean to control whether autopublishing functionality should be on or off (0 #=> false, 1 #=> true)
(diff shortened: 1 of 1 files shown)


app/importers/npr_article_importer.rb
```diff
@@ -64,10 +64,15 @@ def elligible_for_autopublish? npr_story
     end
 
     def auto_publish remote_article, npr_story
-      if elligible_for_autopublish?(npr_story)
+
+      # A boolean to control whether autopublishing functionality should be on or off (0 #=> false, 1 #=> true)
+      npr_auto_publish_on = DataPoint.find_by(data_key: "npr_auto_publish_on").try(:data_value).try(:to_i) || 1
+
+      # Check that all conditions pass: auth-publish is on, the article is elligible, and the environment is either production or test
+      if !npr_auto_publish_on.zero? && elligible_for_autopublish?(npr_story) && (Rails.env == 'production' || Rails.env == 'test')
```
nakedsushi: Consider this for legibility reasons:
`(Rails.env.production? || Rails.env.test?)`

app/importers/npr_article_importer.rb
```diff
@@ -111,7 +114,7 @@ def sync
           existing_story.update headline: remote_article_headline, teaser: npr_story.teaser, url: npr_story.link_for("html")
 
           # If the current npr story was published earlier than our delay period
-          if (Rails.env == 'production' || Rails.env == 'test') && existing_story.published_at < (npr_auto_publish_delay || 120).minutes.ago && existing_story.is_new
+          if existing_story.published_at < (npr_auto_publish_delay || 120).minutes.ago && existing_story.is_new
```
nakedsushi: If this DataPoint was not previously entered in, then `npr_auto_publish_delay` might be nil, which might break this line since `nil.minutes.ago` => `NoMethodError: undefined method `minutes' for nil:NilClass`

app/importers/npr_article_importer.rb
arellaTV: @nakedsushi Got it! Will do, thank you! 🙌

app/importers/npr_article_importer.rb
arellaTV: @nakedsushi  I think in this case, if npr_auto_publish_delay wasn't previously entered, it then defaults to integer `120` before it does the `.minutes.ago` if I'm not mistaken.

Ravenstine: @nakedsushi @arellaTV Just a suggestion... there's already a Settings API, and DataPoints is really a legacy thing that's meant for data that would get pulled in to content.  I'd suggest that, in the future, we either move this setting over to Settings, or we find a non-Outpost settings API for that config to live in.

PR #1248, opened 2018-07-30
Update homepage to include latest 5 LAist entries

arellaTV:
Major changes:

- There is a section at the right rail of the homepage that lists the latest 5 LAist entries
- "Latest NPR headlines" are moved down to position B
- Ad position A is moved outside of the "Latest NPR headlines" block

Note:

- There isn't any deduping logic, so editors have to be mindful of what is in the middle rail/ side rail

### DIFF
--- a/app/helpers/homepage_helper.rb
+++ b/app/helpers/homepage_helper.rb
@@ -40,6 +40,20 @@ module HomepageHelper
   def media_extra content, contents
     render partial: "shared/media/components/extra", locals: {content: content, contents: contents}
   end
+  def latest_from_laist
+    # Perform a get request from the LAist API, and return an empty array if it fails
+    begin
+      response = RestClient.get('http://laist.com/mt/mt-data-api.cgi/v3/sites/1/entries?limit=5')
+      json_response = JSON.parse(response.body)
+      if json_response && json_response["items"]
+        json_response["items"]
+      else
+        []
+      end
+    rescue
+      []
+    end
+  end
   def latest_stories content
     # Takes a collection of any model objects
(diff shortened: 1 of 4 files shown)


app/helpers/homepage_helper.rb
```diff
@@ -40,6 +40,20 @@ def media_meta(feature:, public_datetime:, updated_at:)
   def media_extra content, contents
     render partial: "shared/media/components/extra", locals: {content: content, contents: contents}
   end
+  def latest_from_laist
+    # Perform a get request from the LAist API, and return an empty array if it fails
+    begin
+      response = RestClient.get('http://laist.com/mt/mt-data-api.cgi/v3/sites/1/entries?limit=5')
```
nakedsushi: Instead of hard-coding this URL in, consider adding it as a config variable like the ones in consul's secrets.yml

arellaTV: @nakedsushi Ohh okay, yeah that makes sense! I'll add to secrets and test that out now 🙌

app/helpers/homepage_helper.rb
```diff
@@ -40,6 +40,21 @@ def media_meta(feature:, public_datetime:, updated_at:)
   def media_extra content, contents
     render partial: "shared/media/components/extra", locals: {content: content, contents: contents}
   end
+  def latest_from_laist
+    # Perform a get request from the LAist API, and return an empty array if it fails
+    begin
+      api_base_url = Rails.configuration.x.api.laist.endpoint
```
arellaTV: @nakedsushi 
It should now look for the laist endpoint and apply that as a prefix

PR #1249, opened 2018-07-30
Megaphone integration

arellaTV:
### Major Changes
- Adds `external_podcast_id` (the megaphone id) to the `Podcast` table and its input field in the Podcast edit page in Outpost
- Refactor the redirect functionality so that any `podcast_url` other than ones with `scpr.org/podcasts` are redirected with a `301` status code
- Creates a Megaphone record if one doesn't exist yet
- Updates the Megaphone record when it is updated in Outpost (this time, you can update the `pubdate`, `draft`, `title`, and `summary`)
- Updates the Megaphone record if the audio file is uploaded through Outpost
- Deletes the record from Megaphone when it is deleted from Outpost
- Updates gems and packages because of vulnerabilities

### Tests
- [show_episode_spec](https://github.com/SCPR/SCPRv4/blob/b894c5df538be74232f5573b3e47ae9be84b208a/spec/models/content_base/show_episode_spec.rb)
- [podcasts_controller_spec](https://github.com/SCPR/SCPRv4/blob/b894c5df538be74232f5573b3e47ae9be84b208a/spec/controllers/podcasts_controller_spec.rb)
- [audio_spec](https://github.com/SCPR/SCPRv4/blob/b894c5df538be74232f5573b3e47ae9be84b208a/spec/models/audio_spec.rb)
- [spec_helper](https://github.com/SCPR/SCPRv4/blob/b894c5df538be74232f5573b3e47ae9be84b208a/spec/spec_helper.rb)

I'm so sorry for the big PR! This one was a big feature, so it has several components. You can test it out in staging by adding `ShowEpisodes` to Airtalk and looking at the Megaphone CMS, or this feed: https://feeds.megaphone.fm/airtalk. Note that saving in outpost won't immediately show in the feed until the audio has finished processing and status is set to `Published`.


app/models/show_episode.rb
```diff
@@ -258,14 +262,117 @@ def build_rundown_association(rundown_hash, segment)
     )
   end
 
-  def update_podcast
-    if @podcast_request_body.present?
+  def create_podcast_episode
+    podcast_id = self.try(:show).try(:podcast).try(:external_podcast_id)
+    draft = self.status == 5 ? false : true
```
nakedsushi: Is the tertiary necessary here? Doesn't `self.status == 5` return a boolean anyway?

app/models/show_episode.rb
```diff
@@ -258,14 +262,117 @@ def build_rundown_association(rundown_hash, segment)
...
+    body = {
+      author: self.show.title,
```
nakedsushi: Why is the author attribute set to the show title? Is this how Megaphone wants it set up?

app/models/show_episode.rb
```diff
@@ -258,14 +262,117 @@ def build_rundown_association(rundown_hash, segment)
...
+      draft: draft,
+      externalId: "#{self.obj_key}__#{Rails.env}",
+      pubdateTimezone: Time.zone.name,
+      pubdate: self.air_date || Time.zone.now + 1.year,
```
nakedsushi: Why set the pubdate to 1 year in the future by default?

app/models/show_episode.rb
```diff
@@ -258,14 +262,117 @@ def build_rundown_association(rundown_hash, segment)
...
+    end
+
+    if podcast_id.present? && @podcast_episode_record.nil?
+      begin
+        $megaphone
+          .podcast(podcast_id)
+          .episode
+          .create(body)
+      rescue
+        {}
```
nakedsushi: If this `create_podcast_episode` method were to execute successfully, would it return some sort of associative array or hash object `{something: something}`?  Because if you want to rescue megaphone errors and fail gracefully, whatever the rescue returns should match the type of object that this method should normally return.

app/models/show_episode.rb
```diff
@@ -258,14 +262,117 @@ def build_rundown_association(rundown_hash, segment)
...
+
+    if podcast_id.present? && podcast_episode_record.present?
+      episode_id = podcast_episode_record['id']
+      begin
+        $megaphone
+          .podcast(podcast_id)
+          .episode(episode_id)
+          .delete
+      rescue
+        {}
```
nakedsushi: Same comment here. Does `delete_podcast_episode` return a `{}` type object normally?

app/models/show_episode.rb
```diff
@@ -258,14 +262,117 @@ def build_rundown_association(rundown_hash, segment)
...
+
+  def update_podcast_episode
+    podcast_id = self.try(:show).try(:podcast).try(:external_podcast_id)
+
+    # If a podcast episode doesn't exist on Megaphone's side, try to create it
+    if podcast_id && @podcast_episode_request_body.nil?
+      return create_podcast_episode
+    end
+
+    property_mapper = {
```
nakedsushi: Any reason why we use hashrocket notation here and colons elsewhere like in line 268: `body = { author:...}` ?

app/models/show_episode.rb
arellaTV: @nakedsushi This is actually more related to how `ShowEpisode's` work in that they don't have a changeable byline field and defaults to KPCC, i.e. https://github.com/SCPR/SCPRv4/blob/b894c5df538be74232f5573b3e47ae9be84b208a/app/concerns/concern/associations/bylines_association.rb#L47-L49.

I noticed when you convert an episode to an article, this happens:
https://github.com/SCPR/SCPRv4/blob/b894c5df538be74232f5573b3e47ae9be84b208a/app/models/show_episode.rb#L203

So it seemed like that was what made the most sense.

app/models/show_episode.rb
arellaTV: @nakedsushi Ah! You're right! It seems better written like this:
```
draft = self.status == 5
```

I'll change that in the upcoming commit 🙌

app/models/show_episode.rb
arellaTV: @nakedsushi `pubdate` is a required field for Megaphone, but not required for Outpost, so there was a chance that `self.air_date` would resolve to nil. That was the best solution I could come up with at the time, but maybe I could spend more time to think of another approach

app/models/show_episode.rb
arellaTV: @nakedsushi I don't think there was a particular reason at the time. I think I prefer colon notation, so I can definitely refactor for consistency in the upcoming commit 👍 .

nakedsushi:
:shipit:

arellaTV:
Thanks for the review!

PR #1316, opened 2018-10-05
Changes completion callback from store to cache

arellaTV:
Major change:
- The Megaphone PUT Request waits until after `caching` instead of after `storing` the file

I've tested this on staging with about a week and a half's worth of AirTalk files, and it seems to be working better. A possible reason why this might be more reliable is because the file is supposedly already mature by the time the uploader tries to cache it.

This will hopefully fix the overly long/ randomly restarting episodes happening on AirTalk and Take Two


arellaTV:
Closing as this solution isn't viable

PR #1320, opened 2018-10-16
Changes the GET request to GET from the RSS feed instead of the API

arellaTV:
In conjunction with this Movable Type PR: https://github.com/SCPR/movabletype-laist/pull/16

The changes proposed here will perform a GET request to the RSS feed instead of the API. This is so that the resulting feed doesn't include any `@sponsor` entries, but another nice side effect is that it will most likely be unaffected by the incomplete object bug that occasionally happens with the movabletype data api.

### DIFF
--- a/app/helpers/homepage_helper.rb
+++ b/app/helpers/homepage_helper.rb
@@ -41,13 +41,14 @@ module HomepageHelper
     render partial: "shared/media/components/extra", locals: {content: content, contents: contents}
   end
   def latest_from_laist
-    # Perform a get request from the LAist API, and return an empty array if it fails
+    # Perform a get request to the LAist RSS FEED, and return an empty array if it fails
     begin
-      api_base_url = Rails.configuration.x.api.laist.endpoint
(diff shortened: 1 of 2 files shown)


app/helpers/homepage_helper.rb
```diff
@@ -41,13 +41,14 @@ def media_extra content, contents
     render partial: "shared/media/components/extra", locals: {content: content, contents: contents}
   end
   def latest_from_laist
-    # Perform a get request from the LAist API, and return an empty array if it fails
+    # Perform a get request to the LAist RSS FEED, and return an empty array if it fails
     begin
-      api_base_url = Rails.configuration.x.api.laist.endpoint
```
nakedsushi: Let's make sure to remove this attribute from our secrets store if we don't need to use it after this change.

arellaTV: Ohh okay got it, will do! Thanks @nakedsushi!

PR #1324, opened 2018-10-18
Stop apple news publishing

nakedsushi:
Removes the code that sends blog posts and news articles to Apple News. We haven't been doing it anyway since it only applies to original news content from KPCC.


nakedsushi:
merged via CLI

PR #1333, opened 2018-10-30
Fixes timing issue when sharing a url with Megaphone's API

arellaTV:
## Bug
The resulting file in Megaphone is usually ~1.5x larger than the original file.

## Explanation
Outpost shares the audio url immediately, even if the file itself hasn't finished processing. You can read more about that in this section: https://github.com/SCPR/SCPRv4/blob/0d976b3c003b285f65104f0300a262b35e0fb60c/app/models/audio.rb#L299-L317

I thought I had covered this behavior by adding the following to the `AudioUploader` module (which essentially performs a PUT request after an audio file is stored):
https://github.com/SCPR/SCPRv4/blob/0d976b3c003b285f65104f0300a262b35e0fb60c/app/uploaders/audio_uploader.rb#L11-L40

But when I wrote this, I hadn't considered that even files manually uploaded would still use the url prematurely because it doesn't take into account whether the new audio was recently uploaded or an external url (i.e. resound url):
https://github.com/SCPR/SCPRv4/blob/0d976b3c003b285f65104f0300a262b35e0fb60c/app/models/show_episode.rb#L353-L355

## Proposed Solution
- If the file is coming from our media server, don't perform a PUT request - instead, let the AudioUploader after store callback handle that.
- Add a 2 second delay before performing the PUT request to give some additional padding


app/models/show_episode.rb
```diff
@@ -277,7 +277,7 @@ def create_podcast_episode
 
     available_audio = self.audio.select(&:available?)
 
-    if available_audio.try(:length) > 0
+    if available_audio.try(:length) > 0 && !available_audio.try(:first).try(:url).try(:include?, "media.scpr.org")
```
nakedsushi: A few things here:

1. `.try(:url)` returns a string, which may not respond to `.try(:include?...)`
1. we should extract `media.scpr.org` into a constant since it's also being used in line 353
1. should we just extract all of `available_audio.try(:length) > 0 && !available_audio.try(:first).try(:url).try(:include?, "media.scpr.org")` into an instance variable on the audio object like this?
```
MEDIA_SEVER_DOMAIN='media.scpr.org'

def from_media_server
  self.try(:length) > 0 && !self.try(:first).try(:url).include?(MEDIA_SERVER_DOMAIN)
end
```

app/uploaders/audio_uploader.rb
```diff
@@ -22,19 +22,14 @@ def update_podcast_episode file
...
-            .update({
-              backgroundAudioFileUrl: backgroundAudioFileUrl
-            })
-        rescue
-          {}
-        end
+        background_audio_file_url = File.join(Rails.configuration.x.scpr.audio_url, relative_dir, filename)
+
+        # Enqueue the UpdatePodcastAudio Job 2 seconds from now to give some buffer between processing and uploading
+        Resque.enqueue_at(2.seconds.from_now, Job::UpdateMegaphoneAudio, {
```
nakedsushi: Similar, but `Resque.enqueue_in(2.seconds, Job::...)` might be easier to read

arellaTV:
Squashed and merged with master: a0968237f097d0297c5ec96398394c88463c3b0d

PR #1343, opened 2018-11-06
Adds a check for identical slugs published in the last day before autopublishing

arellaTV:
The changes proposed in this PR add a precaution that should hopefully protect us from weird duplication issues. It flows like this:
- Calculate the eventual slug:
 ```
slug = npr_story.title.parameterize[0...50].sub(/-+\z/, "")
```
- Check the news stories published since yesterday, and see if any of them match the upcoming slug
```
      recent_stories_with_identical_slug = NewsStory
        .where('published_at > ?', DateTime.yesterday)
        .where(slug: slug)
```
- If there is at least one match, disqualify the npr story from being autopublished


arellaTV:
Squashed and merged!

PR #1348, opened 2018-11-09
Fix ad resizing

jo-im:
In this pull request, I fixed the ad resizing so ads won't be cropped in small and medium screen sizes. I also have removed the `ad-resizer.js.coffee` in the `nuevo` folder as it's not being used anymore.


app/assets/javascripts/utilities.js.coffee
```diff
@@ -37,16 +37,19 @@ class scpr.adSizer
...
     resize: (element) ->
         $(element).css
-            "max-width": "100%",
-            "height": "auto"
+            "width": "100%",
+            "height": "100%"
         @removeFixedDimensions(element)
 
+        if (element.nodeName == "AMP-IMG")
+            @setPositionStatic(element)
```
arellaTV: @jo-im Does the `position` property of the `<amp-img>` change when the window resizes? Is that why you need to call `@setPositionStatic` here?

nakedsushi:
Nice, I love PR's that have code deletion 👏

app/assets/javascripts/utilities.js.coffee
jo-im: It's `position: relative` initially which causes the ad to disappear so I have to set the `<amp-img>` to `position: static`. That resize function gets called on load and not on actually resizing the window.

arellaTV: Ohhh okay got it! Thanks for the explanation! 👍

PR #1365, opened 2018-11-30
Add caching to common API methods that get hit from mobile

jo-im:
In this pull request, I added caching to common API methods that get hit from mobile. I have tested my changes locally and on staging by using `Rails.cache.read` and received a cached response. When I added caching to the settings method, the `each` method in `settings/index.json.jbuilder` broke because `@settings` was undefined. I added a change so if `@settings` was `undefined` then it will be set to an empty array.


app/controllers/api/public/v3/schedule_occurrences_controller.rb
```diff
@@ -18,10 +18,12 @@ class ScheduleOccurrencesController < BaseController
     #---------------------------
 
     def index
-      @schedule_occurrences = ScheduleOccurrence.block(@start_time, @length)
-      @pledge_drive = false
-      @display_pledge_status = params[:pledge_status]
-      respond_with @schedule_occurrences
+      Rails.cache.fetch('/api/v3/schedule', expires_in: 5.minutes) do
+        @schedule_occurrences = ScheduleOccurrence.block(@start_time, @length)
+        @pledge_drive = false
```
nakedsushi: This and line 24 don't need to come from cache since they're not "heavy" computational-wise. It's ok to move them above line 21, outside of the cache block.

app/controllers/api/public/v3/settings_controller.rb
```diff
@@ -2,14 +2,16 @@ module Api::Public::V3
...
 
     def index
-      @context  = [params[:context], 'global']
-      @settings = Setting.where(context: @context).where.not(context: nil).to_a
-      @pledge_drive = PledgeDrive.happening.order("starts_at DESC").first
-      if @pledge_drive
-        @settings.unshift @pledge_drive.to_setting
-        @settings.uniq!
+      Rails.cache.fetch("/api/v3/settings/#{params[:context]}", expires_in: 5.minutes) do
+        @context  = [params[:context], 'global']
```
nakedsushi: Same here. Let's take this line out of the cache block and put it above it since it's just setting a variable and doens't require a db lookup.

app/controllers/api/public/v3/schedule_occurrences_controller.rb
jo-im: Oh okay, will move them outside of the cache block. Thank you!

app/controllers/api/public/v3/settings_controller.rb
jo-im: Oh okay, that makes sense! Will move this also above the cache block, thank you!

PR #1374, opened 2018-12-05
 Adds tag level targeting to article and segment pages

arellaTV:
This PR proposes to add tag level targeting to article pages. Similar to category level targeting, but instead passes an array of the tags to google dfp which will allow the ad ops dev to target pages with a specific tag. If there are no tags for a specific page, it defaults to sitewide position targeting.

Documentation:
- Add docs explaining how the ad config scripts work
- Adds a line about running SCPRv4 locally (needs a specific version of mysql installed locally to natively build the mysql gem)
- Removes an intermittently failing test


app/views/shared/ads/_dfp_script_config.html.erb
```diff
@@ -4,6 +4,12 @@
       DFP_CATEGORY = "<%=category.slug%>";
     <% end %>
 
+    <% if story ||= nil %>
```
nakedsushi: If there's no`story` then nothing in this block should run, right? Why not use `if story.present?`

arellaTV: @nakedsushi, Thanks for the review! Yes you're right, the block doesn't run if there isn't a story. I had originally written it that way to match how the category block conditional is set up, but I realize that `if story.present?` is more readable, so I'll go ahead and change to that now. 🙌

arellaTV:
Squashed and merged

PR #1377, opened 2018-12-07
Updates Latest News XML feed for Google Assistant

arellaTV:
Main features:
- Adds an `<image></image>` tag to hold the show icon
- Adds a filter to prevent any NPR content from being inserted into the feed

The following snippet is a new addition to the resulting query. It excludes NPR entries by restricting bylines that look like this: `Julie Rovner | NPR`.
```
"not": {
	"regexp": {
		"byline": "NPR"
	}
}
```
Unfortunately I couldn't directly use the `NewsStory` `:source` attribute as it wasn't one of the properties being indexed in ElasticSearch (and also doesn't exist anyway for `BlogEntries` and `ShowSegments`), but it seems like every time an NPR article is published it includes an `| NPR` suffix at the end of its byline, so I think this will work reliably.

Alternatively, I tried using `term` and `match` instead of `regexp` as they would have been more performant, but neither worked. My closest guess as to why `term` might not have worked in this context was because of the way `byline` is mapped. Although using a regular expression here is less performant, I think the benefit is worth the tradeoff as this function should only execute every time the podcast cache expires  (which is every hour).

Update:
I also added a spec that tests expectations here:
https://github.com/SCPR/SCPRv4/blob/5fbb16ac45f853fcad857c65b9f89d202bc84981/spec/models/podcast_spec.rb#L77-L104


arellaTV:
Thank you for the review! Squashed and merged

PR #1378, opened 2018-12-07
Fixes matching mechanism

arellaTV:
So it looks like the previous `regexp` query was giving me false positives locally. I then realized it's because the `regexp` query will work on unanalyzed fields, and `byline` is mapped as a `string` which is an analyzed field by default.

That means that we can use the `match` query after all. The reason it didn't work before was because I messed up on the shape of the query. This new PR was tested locally and on staging and prevents NPR content from being included in the feed. It also has the added effect of being more performant than the regexp query.


arellaTV:
Thank you for the review! Squashed and merged

PR #1379, opened 2018-12-10
Safely handle nil records in outpost and homepage tag cluster

arellaTV:
The tag cluster on the homepage is failing, blocking the whole `right_asides` from updating (that's why there are stale laist articles in here). The changes here do two things:

1. Removes the Latest LAist headlines out of the right aside block so that it stays fresh
2. Adds a nilcheck to the `tag_cluster` template and outpost aggregator.

The reason why the tag cluster is failing is because one of the `outgoing_references` is resolving to `nil`. The frontend `tag_cluster` template doesn't have a nilcheck, so it raises an error and doesn't complete the cache update when it's supposed to. Outpost aggregator also doesn't have a check for if any of the outgoing refs are `nil`, so it breaks as well (if you go to the Tag edit page in outpost, it will raise an error and not let you edit it).

This probably happened because something that the list used to reference not longer exists (possibly due to deletion), so there's a single nil object in the array (which doesn't have any of the necessary methods).


app/views/better_homepage/_tag_cluster.html.erb
```diff
@@ -3,14 +3,16 @@
   <ul class="c-list c-list--horiz">
     <% omit ||= [] %>
     <% tag.featured_content(omit).each do |feature| %>
-      <div class="o-media o-media--horizontal o-media--sm">
-        <div class="o-media__figure o-media__figure--four-by-three">
-          <img class="o-media__img" src="<%= begin feature.asset.asset.thumb.url rescue '/static/images/fallback-img-square.png' end %>">
+      <% if feature.present? %>
```
nakedsushi: Ideally, it's better to fail early so instead of wrapping this all in an `if` block, how about something like:
```
<% tag.featured_content(omit).each do |feature| %>
   <% next if feature.nil? %>
   <div class="o-media....>
```

arellaTV: Ah! That's a lot better. Thanks for the review @nakedsushi!

PR #1388, opened 2018-12-19
Fix cache api

jo-im:
In this pull request, I added caching to all common API methods that are hit from mobile. The caching is in both the controller and the json builders to ensure quicker responses. I got the average of the response time of each API method before and after caching in SCPRv4 staging and got the percentage of reduced response time: 

- `/api/v3/schedule`: 
   - Average response time before caching: 757ms
   - Average response time after caching: 332ms
   - Response time reduced by: 56%
- `/api/v3/programs`:
   - Average response time before caching: 380ms
   - Average response time after caching: 119ms
   - Response time reduced by: 68%
- `/api/v3/articles`:
   - Average response time before caching: 147ms
   - Average response time after caching: 90ms
   - Response time reduced by: 38%
- `/api/v3/lists`:
   - Average response time before caching: 835ms
   - Average response time after caching: 298ms
   - Response time reduced by: 64%
- `/api/v3/settings`:
   - Average response time before caching: 145ms
   - Average response time after caching: 58ms
   - Response time reduced by: 60%
- `/api/v3/settings/ios`:
   - Average response time before caching: 381ms
   - Average response time after caching: 67ms
   - Response time reduced by: 82%
- `/api/v3/settings/android`:
   - Average response time before caching: 141ms
   - Average response time after caching: 57ms
   - Response time reduced by: 59%

Some issues I encountered was that the caching was not being implemented correctly in `/api/v3/articles` because each article had a `singleton` method so I had to strip it away before responding with it in its controller. Here's a link that explain in more details about stripping `singleton` methods: 
https://stackoverflow.com/questions/6391855/rails-cache-error-in-rails-3-1-typeerror-cant-dump-hash-with-default-proc/6392704#6392704

Here's the link on removing `singleton` methods:
https://stackoverflow.com/questions/23734582/how-to-programmatically-remove-singleton-information-on-an-instance-to-make-it

The other issue was when using `.where` method because it returns the scope and not the actual query. So I attached the `.to_a` method so an array of values are being returned instead of relations. Here is the link that explains more about the problems with caching relations:
https://stackoverflow.com/questions/11218917/confusion-caching-active-record-queries-with-rails-cache-fetch

I have tested on staging SCPRv4 by using `Rails.cache.read('cache-key')` and successfully received the caching. I have also made sure that the cache was missed after every 5 minutes.

### DIFF
--- a/app/controllers/api/public/v3/lists_controller.rb
+++ b/app/controllers/api/public/v3/lists_controller.rb
@@ -2,8 +2,9 @@ module Api::Public::V3
   class ListsController < BaseController
 
     def index
-      @lists = Rails.cache.fetch("/api/v3/lists/#{context}", expires_in: 5.minutes) do
-        if !context.empty?
+      @context = context
+      @lists = Rails.cache.fetch("/api/v3/lists/#{@context}", expires_in: 5.minutes) do
+        if !@context.empty?
           List.visible
             .where("FIND_IN_SET(?, context)", context)
             .order('position ASC').to_a
--- a/app/controllers/api/public/v3/lists_controller.rb
+++ b/app/controllers/api/public/v3/lists_controller.rb
@@ -2,14 +2,14 @@ module Api::Public::V3
   class ListsController < BaseController
 
     def index
-      @lists = Rails.cache.fetch('/api/v3/lists', expires_in: 5.minutes) do
+      @lists = Rails.cache.fetch("/api/v3/lists/#{context}", expires_in: 5.minutes) do
         if !context.empty?
           List.visible
             .where("FIND_IN_SET(?, context)", context)
-            .order('position ASC')
+            .order('position ASC').to_a
         else
           List.visible
-            .order('position ASC')
+            .order('position ASC').to_a
         end
       end
       respond_with @lists
--- a/app/controllers/api/public/v3/lists_controller.rb
+++ b/app/controllers/api/public/v3/lists_controller.rb
@@ -2,13 +2,15 @@ module Api::Public::V3
   class ListsController < BaseController
 
     def index
-      if !context.empty?
-        @lists = List.visible
-          .where("FIND_IN_SET(?, context)", context)
-          .order('position ASC')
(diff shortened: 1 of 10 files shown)


app/views/api/public/v3/articles/index.json.jbuilder
```diff
@@ -2,6 +2,8 @@
 # break anything. The benchmarks showed a ~14% performance boost.
 json.partial! api_view_path("shared", "meta")
 
-json.articles do
-  json.partial! api_view_path("articles", "collection"), articles: @articles
+json.cache! ['/api/v3/articles', @query, @classes, @limit, @page, @conditions], expires_in: 5.minutes do
```
nakedsushi: Since the api/articles_controller already has caching, do we need to cache at the view level here? Would this be double-caching?  We probably only need to do one or the other, right?

jo-im: Oh okay! I can get rid of the caching in the a `articles`  view level. Would you also like me to get rid of all the caching in all the view templates since all the controllers have caching already? Thanks Louise!

app/controllers/api/public/v3/lists_controller.rb
```diff
@@ -2,13 +2,16 @@ module Api::Public::V3
...
 
     def index
-      if !context.empty?
-        @lists = List.visible
-          .where("FIND_IN_SET(?, context)", context)
-          .order('position ASC')
-      else
-        @lists = List.visible
-          .order('position ASC')
+      @context = context
```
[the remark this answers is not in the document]
jo-im: Yes, `@context` variable is being used in the lists view template. Would you like me to get rid of the caching in the `lists` view template since it's double-caching?

app/views/api/public/v3/articles/index.json.jbuilder
jo-im: Thank you! I'll make another branch and test the speed again, thanks so much Louise!

PR #1390, opened 2018-12-27
Adds better caching for news story pages

arellaTV:
It looks like `NewsController#story` pages are performing the worst. There are two factors involved:
1. AssetHost returns asset jsons a bit slowly
2. Many MySQL calls that are being invoked

This PR proposes to cache the ~10-13 mysql calls performed with each get to the story controller.
Before pushing this branch, the average transaction time hovered around 1,640ms for a 30 minute period of non-stop news story visits by a single client.

screen shot 2018-12-27 at 2 50 28 pm

After the branch was pushed, the average transaction time hovered around 153ms for a 12 minute period of non-stop news story visits by a single client.

screen shot 2018-12-27 at 2 53 16 pm

This isn't an immensely clear result and needs more testing, but I think it's still a good idea to put it up for review to catch any logic issues. I'll update this page with a better analysis of the after in a bit.

**Update:**
After close to 30 minutes of testing, it looks like the initial calls are still a little expensive, but the extra caching blocks should ease it considerably for already cached articles.

screen shot 2018-12-27 at 3 37 04 pm

The above screenshot spiked because of a `GET` request o `a.scpr.org` (the version up on staging isn't rebased, so it doesn't have the caching fix that @nakedsushi implemented in the latest master)


app/controllers/news_controller.rb
```diff
@@ -16,7 +16,13 @@ def story
 
     @content_params[:exclude] = @story
 
-    @category = @story.category
+    @category = Rails.cache.fetch("news_stories/#{@story.cache_key}/category") do
```
nakedsushi: where does `.cache_key` come from?

app/views/layouts/application.html.erb
```diff
@@ -34,9 +34,11 @@
     <iframe src="https://www.googletagmanager.com/ns.html?id=GTM-585PL29" height="0" width="0" style="display:none;visibility:hidden"></iframe>
     <div id="global-pushdown"></div>
     <header class="<%= content_for :header_class %>">
-      <%= cell :masthead %>
-      <% if alert = BreakingNewsAlert.latest_visible_alert %>
-        <%= cell :breaking_news_alert, alert %>
+        <% cache(["masthead", "v1"], expires_in: 1.hour) do %>
```
nakedsushi: If there's breaking news, then it could be lagged for 1 hour because of this cache, right?

PR #1401, opened 2019-01-14
Removes programs with the "online" status from sitemaps

arellaTV:
So I noticed that Repeat and The Big One have the value, "Online" for status. They appear to be the only two to use this value, so if we narrow the scope of the call where sitemaps are generated, both of them should be excluded. Tested to work correctly on my development server.


arellaTV:
Squashed and merged!

PR #1402, opened 2019-01-14
Remove parse init

jo-im:
In this pull request, I removed `Parse` init and `Farse` init scripts. I also removed the `Parse` dependency, `Farse` dependency, and the `create_from_parse` method in the `Member` model since `Parse` and `Farse` are no longer being used.


jo-im:
@nakedsushi Hello Louise, I was able to test it on staging today and I also checked New Relic. The site is still up but I'll test more tomorrow just in case. Thank you so much for the review!

PR #1419, opened 2019-03-06
Fixes duplication issue when scheduling episode for future publish

arellaTV:
## The bugs:
1. When scheduling an episode to be published in the future, outpost creates a duplicate megaphone record instead of updating the existing one.
2. Sometimes ad locations, `pre_count`, `post_count`, and `insertion_points`, don't stick after saving in outpost.

## Explanation:
1. Because the system updates the `status` from `draft` to `published` when an episode is scheduled, the instance variable `@podcast_episode_record` is nil because it was never initialized in its current context. Because of that, every time the system changes the `status`, it would invoke the `create_podcast_episode` function.

2. I wasn't able to recreate the 2nd bug, but it could be due to a missing audio file on Megaphone's side.

## Proposed Solutions:
1. Instead of using the instance variable, `@podcast_episode_record`, we use the class method, `podcast_episode_record` which gives us the same data anyway (by performing a `GET` for that episode record from Megaphone). We also use `.blank?` instead of `.nil?` because all Megaphone requests are rescued with `{}`, and `{}.blank?` and `nil.blank?` both resolve to `false` instead of `true` (which is what we want).
2. I've made it so that you can't update the ad location fields unless the audio file exists on Megaphone first.

## Notes:
- Updated the specs for better clarity
- Updated the fixture to include an `audioFile` property
- Added an empty response stub
- Removed the check for if an audio file is from the media server. This was preventing megaphone from picking up re-uploaded audio. This was meant to protect against early uploads that cause elongated files on megaphone's side, but I think I'd like to see how often it happens with the `megaphoneValidator` lambda and CloudWatch.


arellaTV:
Thanks for reviewing! I updated the helper text a bit, squashed, and merged with master

PR #1432, opened 2019-03-28
Removes shortlist appeals

arellaTV:
The changes proposed in this PR remove Short List sign up boxes.

Note:
- After pushing up to production, we should also change the scroll depth script on GTM to point to the `#comments` element instead of the the newsletter sign up box (since that box won't be in the DOM tree anymore). I think the scroll depth function shown in `app/assets/javascripts/new/single.js` isn't actually in use, so it should definitely be shed once I get around to SCPRv4 clean-up work.


arellaTV:
Squashed, merged, and shipped

PR #1447, opened 2019-05-15
Adds some missing tracking functionality to livestream web client

arellaTV:
#### Major changes:
- Switches the url to triton's test url
- Renames `touchImpressions` to `prerollEndImpressions` since those impressions are sent when the preroll finishes playing
- Adds support for `StaticResource` Companion Ads, which are just images with clickthrough links.
- Selects the correct size from a list of companion ads (won't try to render a 150x600 for example)
- Tracks start, quartiles, and end


app/assets/javascripts/listen_live.js.coffee
```diff
@@ -232,26 +235,37 @@ class scpr.ListenLive
...
+                    el?.html link
+                    el.css(margin:"0 auto").width(c._width)
+                    _submitViewEvent(c)
+              }
               {
                 name: 'HTMLResource'
                 render: (c) ->
                     el?.html(c.HTMLResource.toString())
                     el.css(margin:"0 auto").width(c._width)
+                    _submitViewEvent(c)
```
nakedsushi: Wouldn't this mean each time a preroll is played, `_submitViewEvent` would get fired twice, so we'd be counting 2 views per play?

arellaTV: Hi @nakedsushi, thanks for the review! `_submitViewEvent` should only be called once since it takes the first companion ad that satisfies the requirements (`StaticResource, HTMLResource, or IFrameResource`, and `width=300, height=250`) and renders that only.

arellaTV: I think maybe this part made it look like I was invoking `_submitViewEvent` at the top:
https://github.com/SCPR/SCPRv4/blob/f7fc517a11f6819013b1177f5b94e78bd0e331d3/app/assets/javascripts/listen_live.js.coffee#L238

I actually did this to pass off the instance method to a local variable since calling `@_submitViewEvent` inside the `render` of each resource object was the same as calling `this._submitViewEvent` which was undefined in that scope.

arellaTV:
Thanks again for reviewing! I'll close this for now until we finally pull the switch to triton

PR #1453, opened 2019-05-16
Adds LAist to lists

nakedsushi:
A quick and dirty hack to get LAist content into the lists api for the iOS app.

Files of note:
* `import_laist_articles.rb` -- a scheduled job that will run every hour that ingests the MT entries API, makes a fake `ListItem`-like object, stacks them in an array, and writes that array into Rails.cache
* `index.json.jbuilder` -- Builds a different JSON if the list is an LAist list to accommodate for the fake 
* `laist.rb` -- initialize a bunch of constants that we need to get all of the above working. I put the MT entries URL into a secret since I dunno if we should be putting that out in the public or not


app/views/api/public/v3/lists/index.json.jbuilder
```diff
@@ -2,18 +2,30 @@ json.partial! api_view_path("shared", "meta")
...
+    if list.title ==  LAIST_LIST_TITLE
+      json.id            list.id
+      json.title         list.title
+      json.types         list.content_type.split(',') if list.content_type
+      json.context       list.context
+      json.starts_at     list.starts_at
+      json.ends_at       list.ends_at
+      json.created_at    list.created_at
+      json.updated_at    list.updated_at
+      json.items         Rails.cache.read(LAIST_ENTRIES_CACHE) || []
```
arellaTV: Since both sides of the conditional are mostly the same besides `json.items`, would a refactor like this work:
```
if list is an laist list
  json.items rails.cache.read
else
  json.items normal response
end
```

app/jobs/job/import_laist_articles.rb
```diff
@@ -0,0 +1,78 @@
...
+        end
+
+        log "#{json['items'].size} LAist stories found."
+
+        list_items = []
+        json['items'].each_with_index do |item, i|
+          next if item["status"] != "Publish" # Just in case API is not honoring the published status
+          list_items.push(
+            {
+              id: "laist_entry_#{i}",
```
arellaTV: Great idea to add the `laist_entry_` pre-fix to avoid any possible collision errors

app/views/api/public/v3/lists/index.json.jbuilder
nakedsushi: You know, that's what I tried before but I got a type error so I wonder if it's something about the syntax that I'm not following. I'll try again.

arellaTV: Ohh okay! If that doesn't work, then it makes sense to me to keep it the way you originally wrote it. 👍

nakedsushi:
Commited to master:
https://github.com/SCPR/SCPRv4/compare/0a3b99bbdd54...6817f29fbca4

PR #1454, opened 2019-05-20
Adds correct query for "American Homefront"

arellaTV:
Had a conversation with Adam from WUNC and he showed me how to query for the articles we need. You can see the query in action here:
https://support.pmp.io/search?advanced=1&searchsort=date&text=%22American%20Homefront%22&profile=story

The guid we were provided was for an individual story. So instead of using that, the best way to get these articles is by using the keyword, `"American Homefront"`. The double quotes are needed so that the whole keyword can be used as a query, rather than just `American Homefront` which will separately search "American" and "Homefront" (which will yield far more mostly unrelated articles).


arellaTV:
Thank you for reviewing!

arellaTV:
Squashed and merged with prod

PR #1455, opened 2019-05-20
Remove newsroom features

arellaTV:
### Major changes:
- Removes calls  to the `Newsroom::Client`  from `import_remote_article` and `sync_remote_articles`


arellaTV:
Squashed and merged with master

PR #1463, opened 2019-05-30
Initializes 20 second request timeout for ES operations

arellaTV:
This error was traced a few times which seems to be clogging the queue:
`Faraday::Error::TimeoutError: Operation timed out after 5001 milliseconds with 0 bytes received`

It happens when we interface with Elasticsearch, usually after a `contentbase.query`.

#### Major changes:
- [x] Adds a 20 second request timeout for ES operations

#### Related Docs (from current versions used in SCPRv4):
- https://github.com/elastic/elasticsearch-ruby/tree/1.x/elasticsearch-transport#transport-implementations
- https://www.rubydoc.info/gems/patron/0.4.8/Patron/Session#timeout-instance_method


arellaTV:
Thank you for reviewing! Squashed and merged

PR #1475, opened 2019-07-03
Fixes number of arguments exception

arellaTV:
Melissa reported that pressing the Import button in the Remote Articles section of outpost doesn't work. Checking resque, this failed job is logged whenever trying to import:

Screen Shot 2019-07-03 at 2 35 32 PM

The changes presented in this PR address the wrong argument exception and lets it continue executing.


app/jobs/job/import_remote_article.rb
```diff
@@ -16,7 +16,7 @@ def perform(id, import_to_class)
         end
       end
 
-      def on_failure(error)
+      def on_failure(error, id, import_to_class)
```
nakedsushi: Since we're not doing anything on error with the `error` or any of the arguments, no need to be so specific. We can just do:

`def on_failure(*args)` which will take in any amount of arguments.

arellaTV: Ah! Thank you for the suggestion, that works out a lot better

arellaTV:
Fixed in another PR

PR #1482, opened 2019-07-22
Reenable external programs

arellaTV:
Major changes:
- Allows any number of args on failure handling in asset import (old commit)
- Re-enables external program syncing (keeps remote article syncing disabled)


arellaTV:
squashed/merged

PR #1483, opened 2019-07-22
Add external asset support

arellaTV:
## Major changes:
- Adds `external_asset` column to the `assethost_contentasset` table
- Builds an external asset whenever an npr article is imported using the asset host fallback as a base
- Checks if `external_asset` is present first before performing a `GET` request to a.scpr.org
- Caches the `external_asset` with rails cache and reads from that first before parsing the asset.
- Tests expectations with unit tests

An external asset is built from a fallback, so it responds correctly to all Rails attributes (`small.url`, `eight.url`, etc.). Since this is an external asset and only exists in one cutout, each size attribute will have the same url. It's up to the client to properly render the asset.

External assets are currently only created when an NPR article is imported. You can test this functionality by importing an NPR article in outpost in staging.

## Here are some examples showing the asset in different contexts:
- https://scprv4-staging.scprdev.org/news/2019/07/23/86287/becoming-superman-chronicles-the-life-and-career-o/
- https://scprv4-staging.scprdev.org/environment/
- https://scprv4-staging.scprdev.org/arts/
- https://scprv4-staging.scprdev.org/

## Considerations for version 2 or beyond:
- Add support for external assets in the AssetHost picker
- Add support for attaching an external asset to an article in outpost (I need to check if this is a valid use-case)


arellaTV:
Modified my approach above so that it's more flexible to changes in the scheme/shape of assets. Before we were storing a whole asset baked as a json string into the `external_assets` column. Now we're storing a simple hash that contains common sense properties (title, caption, url).

## Most important changes are in these two files:

https://github.com/SCPR/SCPRv4/blob/8e9e5c04d63f4481c23c8c85e0fc5b15f5621d98/app/importers/npr_article_importer.rb#L263-L275

https://github.com/SCPR/SCPRv4/blob/8e9e5c04d63f4481c23c8c85e0fc5b15f5621d98/app/models/content_asset.rb#L38-L79

arellaTV:
Thanks for the review! Squashed/merged

